PR #31, opened 2019-06-26
Feature/components for faq page Tabulation+InlineAlert

pppemo:


packages/react/src/components/Tabulation/README.md
```diff
@@ -0,0 +1,83 @@
+## The component
```
ugogo:
```suggestion
## Tabulation component
```

packages/react/src/components/Tabulation/index.js
```diff
@@ -0,0 +1,65 @@
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import cx from "classnames";
+
+class Tabulation extends Component {
+  constructor(props) {
```
ugogo: What about using hooks here to simplify the overall codebase?

packages/react/src/components/Tabulation/index.js
```diff
@@ -0,0 +1,65 @@
...
+    super(props);
+    this.state = {
+      activeTabIndex: null,
```
ugogo:
```suggestion
      activeIndex: null,
```

or

```suggestion
      currentIndex: null,
```

?

ugogo: What about setting `0` as default index so the first element will be selected by default?

packages/react/src/components/Tabulation/index.js
```diff
@@ -0,0 +1,65 @@
...
+  };
+
+  render() {
+    const { elements, buildTabLabel } = this.props;
+    const { activeTabIndex } = this.state;
+
+    return (
+      <div className="Tabulation">
+        {elements.map((element, index) => (
+          <span key={element} onClick={() => this.handleTabClick(index)}>
```
ugogo: This should be a `<button />`

packages/react/src/components/Tabulation/index.js
```diff
@@ -0,0 +1,65 @@
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import cx from "classnames";
+
+class Tabulation extends Component {
```
ugogo: What about renaming this component in `Tabs`?

packages/css/src/components/ListItem.css
```diff
@@ -0,0 +1,81 @@
+.ListItem {
+  padding: var(--space-l);
+  position: relative;
+  display: flex;
+  display: -webkit-flex;
+  flex-direction: row;
+  align-items: center;
+  min-width: 320px;
```
ugogo: Shouldn't this be `100%`?

packages/css/src/components/ListItem.css
```diff
@@ -0,0 +1,81 @@
...
+}
+
+.ListItem.ListItem--subitem {
```
ugogo: Here and after you can reduce the selector specificity 

```suggestion
.ListItem--subitem {
```

packages/react/src/components/ListItem/index.js
```diff
@@ -0,0 +1,80 @@
...
+  value,
+  leftIcon,
+  rightIcon,
+  hideDivider,
+  onClick,
+}) => (
+  <div
+    onClick={onClick}
+    className={cx({
+      ListItem: true,
```
ugogo: You can move this in the `cx` declaration

`cx("ListItem", {`

packages/react/src/components/ListItem/index.js
```diff
@@ -0,0 +1,80 @@
+import React from "react";
+import PropTypes from "prop-types";
+import cx from "classnames";
+import { Icon } from "..";
+import { ICONS } from "../../constants/";
+
+const ListItem = ({
```
ugogo: Shouldn't we move the `mini` and `sub` into a `size` prop?
Otherwise, you could end with something like `<ListItem mini sub />` (as raised here https://dev.to/selbekk/the-10-component-commandments-2a7f)
It also makes the API more readable

packages/react/src/components/ListItem/index.js
```diff
@@ -0,0 +1,80 @@
...
+  active,
+  title,
```
ugogo: `children`?

packages/react/src/components/ListItem/index.js
```diff
@@ -0,0 +1,80 @@
...
+  subtitle,
+  subitem,
+  miniitem,
+  value,
+  leftIcon,
+  rightIcon,
+  hideDivider,
```
ugogo:
```suggestion
  withDivider,
```
?

packages/react/src/components/ListItem/index.js
```diff
@@ -0,0 +1,80 @@
...
+    })}
+  >
+    {leftIcon && (
+      <div className="ListItem-leftIcon">
+        <Icon icon={leftIcon} />
+      </div>
+    )}
+    <div>
+      <div className="ListItem-title">{title}</div>
+      {!!subtitle && <div className="ListItem-subtitle">{subtitle}</div>}
```
ugogo:
```suggestion
      {subtitle && <p className="ListItem-subtitle">{subtitle}</p>}
```

packages/react/src/components/ListItem/index.js
```diff
@@ -0,0 +1,80 @@
...
+    </div>
+    <div className="ListItem-valueContainer">
+      <span>{value}</span>
+      <span className="ListItem-itemIcon">
+        {(onClick || rightIcon) && (
```
ugogo: Shouldn't be the `onClick` be on the main parent?
So we can have a larger clickable area?

packages/react/src/components/ListItem/index.js
```diff
@@ -0,0 +1,80 @@
...
+    {leftIcon && (
+      <div className="ListItem-leftIcon">
+        <Icon icon={leftIcon} />
+      </div>
+    )}
+    <div>
+      <div className="ListItem-title">{title}</div>
+      {!!subtitle && <div className="ListItem-subtitle">{subtitle}</div>}
+    </div>
+    <div className="ListItem-valueContainer">
```
ugogo: What about moving this into a `rightEnhancer` prop?
So the user can have a larger control on what he wants to display?

packages/react/src/components/ListItem/index.js
```diff
@@ -0,0 +1,80 @@
...
+      <span className="ListItem-itemIcon">
+        {(onClick || rightIcon) && (
+          <Icon icon={rightIcon || ICONS.IconArrowRight} />
+        )}
+      </span>
+    </div>
+  </div>
+);
+
+const ICONS_KEYS = Object.keys(ICONS);
```
ugogo: This declaration should be higher 🙂

packages/react/src/components/InlineAlert/index.js
```diff
@@ -0,0 +1,47 @@
...
+const ALERT_TYPES = Object.values(INLINE_ALERT_TYPES);
+
+const InlineAlert = ({ title, message, type, onClose, ...props }) => (
+  <div className={cx("InlineAlert", `InlineAlert--${type}`)} {...props}>
+    {onClose && (
+      <span className="InlineAlert-iconClose" onClick={onClose}>
+        <Icon icon={ICONS.IconClose} />
+      </span>
+    )}
+    <div className="InlineAlert-icon">
```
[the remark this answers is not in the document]
pppemo: `InlineAlert-icon` is hardly-placing the icon in the right top corner - that's the only case why it's there. I can see that `Icon` is accepting `className` property, would you prefer this classname to be passed like that?

packages/react/src/components/InlineAlert/index.js
```diff
@@ -0,0 +1,47 @@
...
+  [INLINE_ALERT_TYPES.SUCCESS]: ICONS.IconCheck,
+  [INLINE_ALERT_TYPES.ERROR]: ICONS.IconStop,
+};
+
+const ALERT_TYPES = Object.values(INLINE_ALERT_TYPES);
+
+const InlineAlert = ({ title, message, type, onClose, ...props }) => (
+  <div className={cx("InlineAlert", `InlineAlert--${type}`)} {...props}>
+    {onClose && (
+      <span className="InlineAlert-iconClose" onClick={onClose}>
```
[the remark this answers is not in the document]
pppemo: Ok, but why like that?

packages/react/src/components/InlineAlert/index.js
```diff
@@ -0,0 +1,47 @@
...
+        <Icon icon={ICONS.IconClose} />
+      </span>
+    )}
+    <div className="InlineAlert-icon">
+      <Icon icon={ALERT_ICONS[type]} />
+    </div>
+    <div>
+      <div className="InlineAlert-title">{title}</div>
+      <div>{message}</div>
```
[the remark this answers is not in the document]
pppemo: 👍

packages/react/src/components/InlineAlert/index.js
```diff
@@ -0,0 +1,47 @@
...
+    {onClose && (
+      <span className="InlineAlert-iconClose" onClick={onClose}>
+        <Icon icon={ICONS.IconClose} />
+      </span>
+    )}
+    <div className="InlineAlert-icon">
+      <Icon icon={ALERT_ICONS[type]} />
+    </div>
+    <div>
+      <div className="InlineAlert-title">{title}</div>
```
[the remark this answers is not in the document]
pppemo: But styles of `h1` are completely different than these that are expected by design :(

image

image

packages/react/src/components/InlineAlert/index.js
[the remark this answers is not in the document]
ugogo: For an accessibility matter. 
`<button />` are here to triggers actions, where `<span />` doesn't deserve anything but texts from time to time.

packages/react/src/components/InlineAlert/index.js
```diff
@@ -0,0 +1,47 @@
...
+      <div>{message}</div>
+    </div>
+  </div>
+);
+
+InlineAlert.propTypes = {
+  /** Type of alert */
```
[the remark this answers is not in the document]
pppemo: Yup. In a component view, top right corner, if you click "Show Info" these comments are used as descriptions for props documentation. They're not required, but sometimes helpful :)

packages/react/src/components/InlineAlert/index.stories.js
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+import { storiesOf } from "@storybook/react";
+import { withKnobs, select, text } from "@storybook/addon-knobs";
+import { withInfo } from "@storybook/addon-info";
+
+import InlineAlert from "./";
+import { INLINE_ALERT_TYPES } from "../../constants/";
+
+storiesOf("Alerts")
```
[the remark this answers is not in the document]
pppemo: So that was my idea: :)

Since the amount of components is growing, I was thinking about grouping them and putting all alerts into `Alerts`. But maybe it's too early to do that :)

packages/css/src/components/InlineAlert.css
```diff
@@ -0,0 +1,44 @@
+.InlineAlert {
+  position: relative;
+  margin: var(--space-m) 0;
+  padding: var(--space-xl);
+  font-size: var(--font-size--m);
+  line-height: var(--line-height--m);
+  border-radius: var(--space-m);
+  outline: none;
```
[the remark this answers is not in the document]
pppemo: 👍 No idea where it came from :)

packages/css/src/components/InlineAlert.css
```diff
@@ -0,0 +1,44 @@
...
+  border: 0;
+  color: var(--color-white);
+  display: flex;
+  display: -webkit-flex;
```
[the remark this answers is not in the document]
pppemo: Ooh, remembered about it, but wasn't sure if that's the case for Flamingo build as well.
Removing.

packages/css/src/components/InlineAlert.css
```diff
@@ -0,0 +1,44 @@
...
+  flex-direction: row;
+}
+
+.InlineAlert .InlineAlert-title {
```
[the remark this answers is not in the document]
pppemo: 👍

packages/react/src/components/InlineAlert/README.md
```diff
@@ -0,0 +1,28 @@
+## The component
```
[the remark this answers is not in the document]
pppemo: 👍

packages/react/src/components/InlineAlert/README.md
```diff
@@ -0,0 +1,28 @@
...
+This component is an inline alert, that can be used to present some kind of information for the users. It has 3 possible types:
+
+- Information
+- Success
+- Error
+
+## Usage
+
+```jsx
+import { InlineAlert, INLINE_ALERT_TYPES } from "@heetch/flamingo-react";
```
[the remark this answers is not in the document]
pppemo: 👍

packages/react/src/components/Tabulation/README.md
pppemo: 👍

packages/react/src/components/Tabulation/index.js
pppemo: Could be. You good for pushing that for next iteration on the components?

packages/react/src/components/Tabulation/index.js
pppemo: 👍

pppemo: I wouldn't do that as I think it's not a good idea, to impose which tab is opened by default. Maybe there's a case, where nothing would be opened?

But! Your track is good. How about introducing additional prop, which would override that `null` if there's a need to?

packages/react/src/components/Tabulation/index.js
pppemo: I'm in favour of that :)
👍

packages/css/src/components/ListItem.css
pppemo: 🤔 Haven't thought about it. Asking Juan...

pppemo: And he is `yes` :)
Changing!

packages/css/src/components/ListItem.css
pppemo: 👍

packages/react/src/components/ListItem/index.js
pppemo: 👍

packages/react/src/components/ListItem/index.js
pppemo: 👍

packages/react/src/components/InlineAlert/index.js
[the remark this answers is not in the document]
ugogo: CSS is your friend 😁

packages/react/src/components/Tabulation/index.js
ugogo: If you're not familiar enough with hooks, I can handle it (if it helps 🙂)

packages/react/src/components/ListItem/index.js
pppemo: 👍

packages/react/src/components/ListItem/index.js
pppemo: I wanted the divider to be on by default. If there's 10 elements, only the last one will be with `hideDivider`

packages/react/src/components/ListItem/index.js
pppemo: `!!` ✅ 
`<p>` messes with my styles :/

packages/react/src/components/ListItem/index.js
pppemo: The `onClick` itself is bind to the main top `<div>` - indeed
Here it's only to check if `onClick` is there. If so - it's displaying the `>` icon

packages/react/src/components/ListItem/index.js
pppemo: I was thinking about it, but in the end I thought that we shouldn't give user that control. They should be only allowed to show what the component allows them to do - that is aligned with Flamingo. If we allow them to inject whatever they want to - we're screwed :/

packages/react/src/components/ListItem/index.js
pppemo: 👍

pppemo:
And regarding the `Tabs` component. My understanding was that all we want here is the tab selector. In this case I'd opt-in for the current implementation - labels are always going be short strings, so an Array seems to be good.
Do you feel it's a good idea to scale up this component and include also the content of pages that are being chosen? I'm not that convinced :/

packages/react/src/components/ListItem/index.js
ugogo: CSS to the rescue 😉

ugogo:
I agree that including the content might be too much for now.
But it doesn't feel right to pass a set of strings as "elements", what if you need to add an Icon or even an HTML element (`<br />`, `<span />`)?

We could go for

```jsx
<Tabs activeIndex={0}>
    <TabItem>Tab 1 title</TabItem>
    <TabItem>Tab 2 title</TabItem>
    <TabItem>Tab 3 title</TabItem>
  </Tabs>
```

wdyt?

packages/react/src/components/InlineAlert/index.js
```diff
@@ -0,0 +1,49 @@
...
+  [INLINE_ALERT_TYPES.SUCCESS]: ICONS.IconCheck,
+  [INLINE_ALERT_TYPES.ERROR]: ICONS.IconStop,
+};
+
+const ALERT_TYPES = Object.values(INLINE_ALERT_TYPES);
+
+const InlineAlert = ({ title, children, type, onClose }) => (
+  <div
+    className={cx("InlineAlert", `InlineAlert--${type}`, {
+      "is-closeable": !!onClose,
```
ugogo:
```suggestion
      "is-closeable": onClose,
```

packages/react/src/components/InlineAlert/index.js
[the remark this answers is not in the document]
ugogo: If we can reduce the markup, I'd go for it, wdyt?

packages/react/src/components/InlineAlert/index.js
```diff
@@ -0,0 +1,49 @@
...
+  >
+    {onClose && (
+      <span className="InlineAlert-iconClose" onClick={onClose}>
+        <Icon icon={ICONS.IconClose} />
+      </span>
+    )}
+    <div className="InlineAlert-icon">
+      <Icon icon={ALERT_ICONS[type]} />
+    </div>
+    <div>
```
ugogo: Is this empty `<div />` needed?

packages/react/src/components/InlineAlert/index.js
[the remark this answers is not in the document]
ugogo: Oh sweet! 🙂

packages/react/src/components/InlineAlert/index.stories.js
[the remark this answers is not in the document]
ugogo: What suits you the best 👌

packages/react/src/components/InlineAlert/index.stories.js
```diff
@@ -0,0 +1,27 @@
+import React from "react";
+import { storiesOf } from "@storybook/react";
+import { withKnobs, select, text } from "@storybook/addon-knobs";
+import { action } from "@storybook/addon-actions";
+import { withInfo } from "@storybook/addon-info";
+
+import InlineAlert from "./";
+import { INLINE_ALERT_TYPES } from "../../constants/";
+
+storiesOf("Alerts")
```
ugogo: According to the Notion, there should be:
- A story with all states
- A `Playground` story

packages/react/src/constants/index.js
```diff
@@ -40,3 +40,15 @@ export const INTENTS = {
...
   TERTIARY: "tertiary",
 };
+
+export const INLINE_ALERT_TYPES = {
+  INFORMATION: "information",
+  SUCCESS: "success",
+  ERROR: "error",
+};
+
+export const LIST_ITEM_TYPES = {
```
ugogo:
```suggestion
export const LIST_ITEM_SIZES
```
?

packages/css/src/components/InlineAlert.css
```diff
@@ -0,0 +1,45 @@
...
+}
+
+.InlineAlert-title {
+  font-weight: bold;
+}
+
+.InlineAlert-icon {
+  padding-right: var(--space-xl);
+  display: flex;
+  display: -webkit-flex;
```
ugogo: Not needed with autoprefixer 😉

packages/css/src/components/InlineAlert.css
```diff
@@ -0,0 +1,45 @@
...
+  top: 5px;
+  right: 10px;
+  cursor: pointer;
+}
+
+.InlineAlert.is-closeable {
+  padding-right: 3rem;
+}
+
+.InlineAlert.InlineAlert--information {
```
ugogo:
```suggestion
.InlineAlert--information {
```

packages/css/src/components/InlineAlert.css
```diff
@@ -0,0 +1,45 @@
...
+  background-color: var(--color-sonic);
+}
+
+.InlineAlert.InlineAlert--success {
```
ugogo:
```suggestion
.InlineAlert--success {
```

packages/css/src/components/InlineAlert.css
```diff
@@ -0,0 +1,45 @@
...
+  background-color: var(--color-link);
+}
+
+.InlineAlert.InlineAlert--error {
```
ugogo:
```suggestion
.InlineAlert--error {
```

packages/react/src/components/InlineAlert/index.stories.js
```diff
@@ -0,0 +1,27 @@
...
+  .addDecorator(withKnobs)
+  .add(
+    "InlineAlert",
+    withInfo("")(() => (
```
ugogo: So this knob if for displaying the comments in the component, right?

packages/react/src/components/ListItem/index.js
```diff
@@ -0,0 +1,90 @@
+import React from "react";
+import PropTypes from "prop-types";
+import cx from "classnames";
+import { Icon } from "..";
+import { ICONS, LIST_ITEM_TYPES } from "../../constants/";
+
+const ICONS_KEYS = Object.keys(ICONS);
+
+const ListItem = ({
+  active,
```
ugogo:
```suggestion
 isActive,
```

packages/react/src/components/ListItem/index.js
ugogo: I don't think it's gonna be a problem because, in the end, we will also need this flexibility

packages/css/src/components/ListItem.css
```diff
@@ -0,0 +1,81 @@
+.ListItem {
+  padding: var(--space-l);
+  position: relative;
+  display: flex;
+  display: -webkit-flex;
+  flex-direction: row;
+  align-items: center;
+  width: 100%;
```
ugogo: Not needed, a `div` has this by default 😉

packages/react/src/components/ListItem/index.js
```diff
@@ -0,0 +1,90 @@
...
+  value,
+  leftIcon,
+  rightIcon,
+  hideDivider,
+  onClick,
+}) => (
+  <div
+    onClick={onClick}
+    className={cx("ListItem", {
+      "ListItem--subitem": type === LIST_ITEM_TYPES.SUB,
```
ugogo:
```suggestion
      "ListItem--sub": type === LIST_ITEM_TYPES.SUB,
```
No need to repeat `item` here 🙂

packages/react/src/components/ListItem/index.js
```diff
@@ -0,0 +1,90 @@
...
+      "ListItem--miniitem": type === LIST_ITEM_TYPES.MINI,
```
ugogo:
```suggestion
      "ListItem--mini": type === LIST_ITEM_TYPES.MINI,
```

No need to repeat item here too 🙂

packages/react/src/components/ListItem/index.js
```diff
@@ -0,0 +1,90 @@
...
+      "is-active": active,
+      "is-clickable": !!onClick,
+      "ListItem-divider": !hideDivider,
```
ugogo:
```suggestion
      "has-divider": !hideDivider,
```

It's a modifier, not a children 😉

packages/css/src/components/ListItem.css
```diff
@@ -0,0 +1,81 @@
...
+  display: -webkit-flex;
+  flex-direction: row;
+  align-items: center;
+}
+
+.ListItem-itemIcon {
+  width: var(--space-m);
+  margin: 0 var(--space-m);
+  display: flex;
+  display: -webkit-flex;
```
ugogo: Here and in other places in the same file, not needed with autoprefixer

packages/react/src/components/ListItem/index.js
```diff
@@ -0,0 +1,90 @@
...
+    })}
+  >
+    {leftIcon && (
+      <div className="ListItem-leftIcon">
```
ugogo: I think you can pass this class to the `Icon` directly

packages/react/src/components/ListItem/index.js
```diff
@@ -0,0 +1,90 @@
...
+    <div>
+      <div
+        className={cx({
+          "ListItem-title": type !== LIST_ITEM_TYPES.MINI,
+          "ListItem-subtitle": type === LIST_ITEM_TYPES.MINI,
+        })}
+      >
+        {children}
+      </div>
+      {subtitle && <div className="ListItem-subtitle">{subtitle}</div>}
```
ugogo: It might be better to have a `<p>` here, wdyt?

packages/react/src/components/ListItem/index.js
```diff
@@ -0,0 +1,90 @@
...
+    {leftIcon && (
+      <div className="ListItem-leftIcon">
+        <Icon icon={leftIcon} />
+      </div>
+    )}
+    <div>
+      <div
+        className={cx({
+          "ListItem-title": type !== LIST_ITEM_TYPES.MINI,
+          "ListItem-subtitle": type === LIST_ITEM_TYPES.MINI,
```
ugogo: So you could end with 2 `ListItem-subtitle`? 🤔

packages/react/src/components/ListItem/index.js
```diff
@@ -0,0 +1,90 @@
...
+    <div className="ListItem-valueContainer">
+      <span
+        className={cx({
+          "ListItem-title": type !== LIST_ITEM_TYPES.MINI,
+          "ListItem-subtitle": type === LIST_ITEM_TYPES.MINI,
+        })}
+      >
+        {value}
+      </span>
+      <span className="ListItem-itemIcon">
```
ugogo: You can pass this class to the `Icon` directly

packages/react/src/components/ListItem/index.js
ugogo: Oh gotcha 👍

packages/react/src/components/ListItem/index.js
```diff
@@ -0,0 +1,90 @@
...
+        {(onClick || rightIcon) && (
+          <Icon icon={rightIcon || ICONS.IconArrowRight} />
```
ugogo: Shouldn't this be set as a `defaultProp`?

packages/react/src/components/ListItem/README.md
```diff
@@ -0,0 +1,35 @@
+## The component
```
ugogo:
```suggestion
## ListItem component
```

packages/react/src/components/ListItem/index.js
```diff
@@ -0,0 +1,90 @@
...
+        )}
+      </span>
+    </div>
+  </div>
+);
+
+ListItem.propTypes = {
+  active: PropTypes.bool,
+  children: PropTypes.string.isRequired,
```
ugogo:
```suggestion
  children: PropTypes.oneOfTypes([PropTypes.string, PropTypes.node]).isRequired,
```

packages/react/src/components/ListItem/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Usage
+
+```jsx
+import { ListItem, ICONS } from "@heetch/flamingo-react";
+
+export default () => (
+  <>
+    <ListItem
+      title="List Item"
```
ugogo: This has been moved to a `children` prop 🙂

packages/css/src/components/ListItem.css
```diff
@@ -0,0 +1,81 @@
...
+  color: var(--color-tolopea);
+}
+
+@media only screen and (max-width: 460px) {
+  .ListItem {
+    padding: var(--space-xl) var(--space-l);
```
ugogo: How does it react with other sizes?

packages/react/src/components/ListItem/index.stories.js
```diff
@@ -0,0 +1,60 @@
...
+} from "@storybook/addon-knobs";
+import { withInfo } from "@storybook/addon-info";
+import { action } from "@storybook/addon-actions";
+import { ICONS, LIST_ITEM_TYPES } from "../../constants/";
+
+import ListItem from "./";
+
+ICONS.None = null;
+
+storiesOf("List Items")
```
ugogo: Same comment as for `InlineAlert` about the stories organization

packages/css/src/index.css
```diff
@@ -7,8 +7,11 @@
 @import "./components/Heading.css";
 @import "./components/Helper.css";
 @import "./components/Icon.css";
+@import "./components/InlineAlert.css";
 @import "./components/Input.css";
 @import "./components/Label.css";
+@import "./components/ListItem.css";
 @import "./components/Select.css";
+@import "./components/Tabs.css";
```
ugogo: Thanks for following the alphabetical order ❤️

packages/react/src/components/ListItem/index.js
pppemo: Yeah :) But I'm rather like - why `<p>` instead of `<div>`? `<div>` works just as well and I don't have to override `<p>`'s styles :)

packages/react/src/components/ListItem/index.js
pppemo: It's like every API. It should be as exposed as needed. If we allow users to inject whatever they want - there's no point of this whole effort that Juan is doing.

Anyways, I've adjusted this component to accept the type of the value

packages/react/src/components/ListItem/index.js
ugogo: Semantic my friend, semantic 👌

packages/css/src/components/InlineAlert.css
pppemo: 👍

packages/css/src/components/InlineAlert.css
pppemo: 👍

packages/css/src/components/InlineAlert.css
pppemo: 👍

packages/css/src/components/InlineAlert.css
pppemo: 👍

packages/css/src/components/ListItem.css
pppemo: 👍

packages/css/src/components/ListItem.css
pppemo: 👍 replaced all occurences

packages/css/src/components/ListItem.css
pppemo: If you're asking about another viewports - the specs required it only to change on a phone.

If you're asking about other sizes of `ListItem` component (normal, sub, mini) it's been changed now to:

```
.ListItem:not(.ListItem--miniitem):not(.ListItem--subitem) {
```

so that i'ts only applied to the normal item

packages/css/src/index.css
pppemo: Only like that :-)

packages/react/src/components/InlineAlert/index.js
pppemo: 👍

packages/react/src/components/InlineAlert/index.js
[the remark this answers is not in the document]
pppemo: 👍

packages/react/src/components/InlineAlert/index.js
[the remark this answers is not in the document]
pppemo: Tried that, but apparently positioning an SVG element with flex inside `<i>` does not work as expected. Honestly didn't want to loose more time on this. Do you believe it's really worth reworking?

packages/react/src/components/InlineAlert/index.js
pppemo: I'd say so. It's because of how I positioned the elements in containers.
Container with icon to the left, then **this** `<div>` with two values positioned vertically inside

packages/react/src/components/InlineAlert/index.js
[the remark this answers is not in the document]
pppemo: So that's what I'm trying to understand. Of course it's possible to style it, but what's the value of changing it to `h1`?

What I'm thinking is that `div` works just as fine, and when I change it to `h1` there's an additional work I need to do. And this will repeat for all upcoming occurrences :/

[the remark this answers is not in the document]
pppemo: What I'm sensing is that you might want to have it this way here to have it done `by the book` and as everywhere else, so that we maintain this library mind-blowing. Could I be right? :)

If that's the case it's ❤️ but maybe in that particular case it's not worth the effort, hmmm? :P

packages/react/src/components/InlineAlert/index.stories.js
pppemo: Absolutely!

packages/react/src/components/InlineAlert/index.stories.js
pppemo: That's a nice story :)

So this know (when passing a readme file here) adds the readme to "Show info" page of a component. Meaning that the Readme would be available also from the storybook itself, not only within the repository.

But they've recently discovered a bug where it fails to parse triple `\`` that we use to paste code blocks.

So the bug is reportedly fixed and it's in beta now I believe. So I'm temporarily passing an empty string there, waiting for the bugfix to become public, and then we'll upgrade the version of the knob and pass Readmes there :)

packages/react/src/components/ListItem/README.md
pppemo: 👍

packages/react/src/components/ListItem/README.md
pppemo: Sweet catch

👍

packages/react/src/components/ListItem/index.js
pppemo: You good with that?

packages/react/src/components/ListItem/index.js
pppemo: 👍

packages/react/src/components/ListItem/index.js
pppemo: 👍

packages/react/src/components/ListItem/index.js
pppemo: 👍

packages/react/src/components/ListItem/index.js
pppemo: Same as before... it doesn't work well and I don't know if it's really worth fixing that now :/

packages/react/src/components/ListItem/index.js
pppemo: No, why? If it's `mini` we use class `subtitle`. Otherwise it's `title`.

What might be confusing here (and that's what you probably mean)
Is that in `mini` type class `subtitle` is used to show actual title.

packages/react/src/components/ListItem/index.js
pppemo: What value would that add?

packages/react/src/components/ListItem/index.js
pppemo: Nice... but no :P

So `ArrowRight` is expected to be shown only when `onClick` is present.
If we put it in `defaultProps` it would be shown always, unless defined otherwise.
Not sure if I was clear enough?

packages/react/src/components/ListItem/index.js
pppemo: 👍

packages/react/src/components/ListItem/index.stories.js
pppemo: 👍

packages/react/src/components/InlineAlert/index.js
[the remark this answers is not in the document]
ugogo: Yeah let's keep it like that, it's not a deal-breaker 👍

packages/react/src/components/InlineAlert/index.js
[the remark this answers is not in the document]
ugogo: It's worth the effort because it's for an accessibility matter! Otherwise, let's use `div`s and `span`s everywhere and we don't care 🤷‍♂ 

Accessibility is important.
We want our products to be accessible by as many people possible, whatever their condition (under a reasonable bar). And it all starts with this first-level components library.
If it means spending 5 additional minutes to add some CSS to match the designs, yeah we should do it.

packages/react/src/components/ListItem/index.js
ugogo: Yeah, let's keep it this way

packages/css/src/components/ListItem.css
ugogo: Alright 👍 
Let's hope this specificity doesn't bite us back 🙂

packages/react/src/components/ListItem/index.js
ugogo: Yep, got it 👍

packages/react/src/components/ListItem/index.js
ugogo: 👍

packages/react/src/components/InlineAlert/index.js
ugogo: 👍

packages/react/src/components/InlineAlert/index.stories.js
ugogo: 👍

packages/react/src/components/ListItem/index.js
ugogo: Semantic 😉

packages/react/src/components/InlineAlert/index.js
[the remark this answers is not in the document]
pppemo: Accessibility is important - no doubt. How does `<p>` help in that?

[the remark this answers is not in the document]
pppemo: Done!

packages/react/src/components/ListItem/index.js
pppemo: 👍

packages/react/src/components/InlineAlert/index.js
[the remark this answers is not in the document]
ugogo: Did you forget to push your commit? 😛

packages/react/src/components/Tabs/index.js
```diff
@@ -0,0 +1,65 @@
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import cx from "classnames";
+
+class Tabs extends Component {
```
ugogo: The same comment about hooks, there is no reason to ship code we can quickly improve
(Let me know if I can help)

packages/react/src/components/Tabs/index.js
```diff
@@ -0,0 +1,65 @@
...
+        ))}
+      </div>
+    );
+  }
+}
+
+Tabs.propTypes = {
+  /** An array of strings or objects that will make tabs content */
+  elements: PropTypes.arrayOf(
+    PropTypes.oneOfType([PropTypes.string, PropTypes.object])
```
ugogo: What about `PropTypes.node`?

ugogo: IIRC there's a linter rule about avoiding `PropTypes.object` in favor of `PropTypes.shape`
Could you update it?

```
PropTypes.shape({
  foo: PropTypes.something
  bar: PropTypes.something
})
```

packages/react/src/components/Tabs/index.stories.js
```diff
@@ -0,0 +1,21 @@
+import React from "react";
+import { storiesOf } from "@storybook/react";
+import { withKnobs, array } from "@storybook/addon-knobs";
+import { withInfo } from "@storybook/addon-info";
+
+import Tabs from "./";
+
+storiesOf("Navigation/Tabs")
```
ugogo:
```suggestion
storiesOf("Tabs")
```

?

packages/react/src/components/Tabs/index.js
```diff
@@ -0,0 +1,65 @@
...
+  };
+
+  render() {
+    const { elements, buildTabLabel } = this.props;
+    const { activeIndex } = this.state;
+
+    return (
+      <div className="Tabs">
+        {elements.map((element, index) => (
+          <span key={element} onClick={() => this.handleTabClick(index)}>
```
ugogo: Semantic 😉

packages/react/src/components/Tabs/index.js
```diff
@@ -0,0 +1,65 @@
...
+            <div className="Tabs-label">
+              {buildTabLabel ? buildTabLabel(element) : element}
+            </div>
+            <div
+              className={cx("Tabs-marker", {
+                "is-active": activeIndex === index,
```
ugogo: 😍

PR #33, opened 2019-07-04
Upgraded Storybook and implemented some storybook addons

pppemo:


packages/react/.babelrc
```diff
@@ -1,6 +1,6 @@
 {
   "exclude": "node_modules/**",
-  "plugins": [],
+  "plugins": ["@babel/plugin-proposal-class-properties"],
```
pppemo: Eslint was complaining to some arrows when this wasn't there

packages/react/.storybook/config.js
```diff
@@ -9,4 +10,47 @@ function loadStories() {
...
+addParameters({
+  options: {
+    name: "Heetch Flamingo",
+    addonPanelInRight: true,
+    hierarchySeparator: /\//,
+  },
+});
+
+const flamingoViewports = {
+  BreakpointS: {
```
ugogo: Is this documented somewhere?
Shouldn't we use `mobile/tablet/desktop`?

packages/react/.babelrc
ugogo: Why is this needed?

pppemo: Eslint was complaining to some arrows when this wasn't there
Don't have a clear understanding why though...

packages/react/.storybook/config.js
pppemo: https://www.notion.so/heetch/Grid-1e18d12216b040c28d0e146125e7fded#dab7afa3ba2947f6a45a98e59d89d56a

packages/react/.babelrc
ugogo: Weird, okay then 👍🏼

packages/react/.storybook/config.js
ugogo: 👍

PR #35, opened 2019-07-05
Component/button

ugogo:


packages/react/src/components/Button/README.md
```diff
@@ -10,6 +10,16 @@ export default () => <Button>Content</Button>;
...
 ## Props
 
+#### `as`: String or Function
+
+```jsx
+export default () => <Button as="a" />;
+```
+
+```jsx
+export default () => <Button as={Link} />;
```
pppemo: Since `Link` is:
```
const Link = ({ children, to, ...props }) => (
  <a {...props} href={to}>
    {children}
  </a>
```
What is the difference between `a` and `Link`?

ugogo: It's just an example 😉
`Link` can be pulled out from `react-router-dom` for example

pppemo: Ohhhhh yup ok :) that makes sense :)

PR #43, opened 2019-07-26
Feature/component/checkbox

pppemo:


packages/react/src/components/Checkbox/index.js
```diff
@@ -0,0 +1,60 @@
...
+        {sublabel && <p className="Checkbox-sublabel">{sublabel}</p>}
+      </div>
+    </label>
+  ) : (
+    buildInputTag()
+  );
+};
+
+Checkbox.propTypes = {
+  onChange: PropTypes.func,
```
ugogo: I'd go with `isRequired`, wdyt?

packages/react/src/components/Checkbox/index.js
```diff
@@ -0,0 +1,60 @@
...
+  isUndefined: PropTypes.bool,
+  isChecked: PropTypes.bool,
```
ugogo:
```suggestion
  checked: PropTypes.bool,
```

To match the HTML element props

packages/react/src/components/Checkbox/index.js
```diff
@@ -0,0 +1,60 @@
...
+
+const Checkbox = ({
+  isUndefined,
+  disabled,
+  children,
+  isChecked: isDefaultChecked,
+  sublabel,
+  onChange,
+}) => {
+  const [isChecked, setIsChecked] = useState(isDefaultChecked);
```
ugogo: :dancing_panda:

packages/react/src/components/Checkbox/index.stories.js
```diff
@@ -0,0 +1,24 @@
+import React from "react";
+import { storiesOf } from "@storybook/react";
+import { boolean, select, text, withKnobs } from "@storybook/addon-knobs";
+
+import Checkbox from ".";
+
+storiesOf("Form controls/Checkbox", module)
+  .addDecorator(withKnobs)
```
ugogo: I wonder if we could move this into the main stories definition to avoid doing it every time 🤔

packages/react/src/components/Checkbox/index.stories.js
```diff
@@ -0,0 +1,24 @@
...
+  .addDecorator(storiesFn => (
+    <div style={{ display: "flex", justifyContent: "center" }}>
```
ugogo: Isn't it supposed to be done within the main flex decorator?
https://github.com/heetch/flamingo/blob/master/packages/react/.storybook/FlexWrapperDecorator.js

packages/react/src/components/Checkbox/index.stories.js
```diff
@@ -0,0 +1,24 @@
...
+      {storiesFn()}
+    </div>
+  ))
+  .add("Playground", () => (
+    <>
```
ugogo: No need for `<>` here 🙂

packages/react/src/components/Checkbox/index.js
```diff
@@ -0,0 +1,60 @@
+import React, { useState } from "react";
+import PropTypes from "prop-types";
+import cx from "classnames";
+
+const Checkbox = ({
+  isUndefined,
+  disabled,
+  children,
+  isChecked: isDefaultChecked,
+  sublabel,
```
ugogo: What about `helper`, to match other form controls helpers?

packages/react/src/components/Checkbox/index.js
```diff
@@ -0,0 +1,60 @@
...
+    buildInputTag()
+  );
+};
+
+Checkbox.propTypes = {
+  onChange: PropTypes.func,
+  isUndefined: PropTypes.bool,
+  isChecked: PropTypes.bool,
+  disabled: PropTypes.bool,
+  children: PropTypes.string,
```
ugogo: Why would this not be required?
Also, should be `PropTypes.node`

packages/react/src/components/Checkbox/index.js
```diff
@@ -0,0 +1,60 @@
...
+      type="checkbox"
+      onChange={onCheckboxStateChange}
+    />
+  );
+
+  return children ? (
+    <label className="Checkbox-label">
+      {buildInputTag()}
+      <div>
+        {children}
```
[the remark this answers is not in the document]
pppemo: To be checked with Juan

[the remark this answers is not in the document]
ugogo: @JuanGGZ The style is not the same, should we use a `Label` component here?

packages/react/src/components/Checkbox/index.js
```diff
@@ -0,0 +1,60 @@
...
+        {sublabel && <p className="Checkbox-sublabel">{sublabel}</p>}
```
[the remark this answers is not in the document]
pppemo: 👍

packages/react/src/components/Checkbox/index.js
pppemo: --> To use a `safeInvoker`

packages/react/src/components/Checkbox/index.js
pppemo: 👍

packages/react/src/components/Checkbox/index.stories.js
pppemo: 👍

packages/react/src/components/Checkbox/index.stories.js
pppemo: 👍

packages/react/src/components/Checkbox/index.stories.js
pppemo: 👍

packages/react/src/components/Checkbox/index.js
pppemo: 👍

packages/react/src/components/Checkbox/index.js
pppemo: * Remove condition for labels
* ESlint - check config in Pemo's VSCode
* Children not required - ask Juan
* Different content for children - ask Juan
* remove label-has-for

packages/react/src/components/Checkbox/index.js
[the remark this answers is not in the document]
pppemo: https://heetch.slack.com/archives/C2BHK4Q4Q/p1564479896022100

packages/react/src/components/Checkbox/index.js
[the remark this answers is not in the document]
pppemo: I've changed var name, but didn't use `Helper` component, as it was as a matter of fact slightly different

packages/react/src/components/Checkbox/index.js
pppemo: Both changed

packages/react/src/components/Checkbox/index.stories.js
pppemo: Done

packages/react/src/components/Checkbox/index.stories.js
pppemo: https://retool.atlassian.net/browse/FRONT-76

packages/react/src/components/Checkbox/index.js
pppemo: I've left conditions for leaving the labels, because if `Checkbox` was left in `label` without an actual label, the object looked like:

image

Rest Juan covered in Slack thread

packages/react/src/components/Checkbox/index.js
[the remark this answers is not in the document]
ugogo: 👍🏼

packages/react/src/components/Checkbox/index.js
```diff
@@ -0,0 +1,62 @@
...
+      onChange={onCheckboxStateChange}
+    />
+  );
+
+  return children ? (
+    <label className="Checkbox-label">
+      {buildInputTag()}
+      <div>
+        {children}
+        {helper && <p className="Checkbox-sublabel">{helper}</p>}
```
ugogo:
```suggestion
        {helper && <p className="Checkbox-helper">{helper}</p>}
```

packages/css/src/components/Checkbox.css
```diff
@@ -0,0 +1,84 @@
...
+}
+
+.Checkbox-label {
+  display: flex;
+  cursor: pointer;
+  color: var(--color-tolopea);
+  font-size: var(--font-size--m);
+}
+
+.Checkbox-sublabel {
```
ugogo:
```suggestion
.Checkbox-helper {
```

packages/react/src/components/Checkbox/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+Checkboxes allow the user to select one or more options from a set.
+
+## Usage
+
+```jsx
+import { Checkbox } from "@heetch/flamingo-react";
+
+export default () => (
+  <Checkbox disabled sublabel="Used when there is more than one ride">
```
ugogo:
```suggestion
  <Checkbox disabled helper="Used when there is more than one ride">
```

pppemo: 🤦‍♂

PR #44, opened 2019-07-30
Introduce Zeit (now) auto deploy

gyndav:
## Business rationale
In order to ease reviewing when adding new components to Storybook, let's use **Zeit** in order to **automatically deploy each PR** on an isolated environment.

## Changes

* introduce Zeit `now`

### DIFF
--- a/docs/how-to-review-new-components.md
+++ b/docs/how-to-review-new-components.md
@@ -1,0 +1,7 @@
+
+## How-to review new components
+
+Each pull request will be automatically deploy on [Zeit](https://zeit.co) via `now`,
+producting a unique URL such as https://flamingo-git-gyndav-featzeit.heetch-eng.now.sh
+
+Each incremental changes will create a new slug.
--- a/now.json
+++ b/now.json
@@ -1,0 +1,12 @@
+{
+  "name": "flamingo",
+  "version": 2,
+  "builds": [
+    {
+      "src": "now-build.sh",
+      "use": "@now/static-build",
+      "config": { "distDir": "packages/react/build" }
+    }
+  ],
+  "regions": ["cdg1", "dub1", "bru1", "lhr1"]
+}
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ Every package is available through NPM
 
 - [React components](packages/react/README.md)
 - [Standalone CSS](packages/css/README.md)
+- [How-to review new components](docs/how-to-review-new-components.md)
 
 ## License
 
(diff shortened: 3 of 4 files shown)


docs/how-to-review-new-components.md
```diff
@@ -0,0 +1,7 @@
+
+## How-to review new components
+
+Each pull request will be automatically deploy on [Zeit](https://zeit.co) via `now`,
+producting a unique URL such as https://flamingo-git-gyndav-featzeit.heetch-eng.now.sh
```
ugogo: How is this `flamingo-git-gyndav-featzeit` built?
Is this related to the branch's name?

now.json
```diff
@@ -0,0 +1,12 @@
+{
+    "name": "flamingo",
+    "version": 2,
```
ugogo: That's the `now` version, not Flamingo's one right?

now.json
```diff
@@ -0,0 +1,12 @@
...
+    "builds": [
+      {
+        "src": "build.sh",
```
ugogo: Should we wrap this into a `script/` folder?
We might have more of this latter, wdyt?

now.json
gyndav: Yup https://zeit.co/docs/v2/advanced/configuration#version

docs/how-to-review-new-components.md
gyndav: What do you mean? I'm afraid I don't really understand the underlying question there.
If you meant the name, indeed, sounds like `project-name-platform-branch_name`. Yet another alias, that we can prevent. The original one here was https://flamingo-fw5vr9suf.now.sh/

now.json
gyndav: Agreed. Done. ✅

docs/how-to-review-new-components.md
ugogo: Yeah sorry I was talking the naming of this URL
`project-name-platform-branch_name` sounds good 👍

gyndav:
@ugogo I'm not really fond of what the config and build files became with introducing `scripts`dir ...

Platform assumes only **relative path** (not absolute one).
Do you mind if we rollback the `scripts` directory change?

ugogo:
@gyndav Ok if we can rename the `.sh` script to `now.build.sh` (or something like that) 🙂

gyndav:
@ugogo done with `now-build.sh` (sorry that's the Shell default separator 😅 )

ugogo:
Not the best but we can live with that 🙂 
Thanks for the help @gyndav!

pppemo:
Cool is this good to be merged now @gyndav ?

gyndav:
for sure @pppemo 🔥

PR #45, opened 2019-07-30
Uploaders/support multiple files

ugogo:


packages/react/src/components/FileUploader/index.js
```diff
@@ -25,45 +25,54 @@ const texts = {
...
-    setInnerFile(eventFile);
-    safeInvoke(onChange(eventFile));
+    setFiles(nextFiles);
+    safeInvoke(onChange(inputFiles));
   };
 
-  const handleResetState = () => {
-    setInnerFile(undefined);
+  const handleDeleteFile = fileToDelete => {
+    const filteredFiles = files.filter(file => file.name !== fileToDelete.name);
```
pppemo:
```suggestion
    const remainingFiles = files.filter(file => file.name !== fileToDelete.name);
```
?

No strong opinion though

packages/react/src/components/FileUploader/index.stories.js
```diff
@@ -19,13 +18,18 @@ stories.add("Playground", () => (
     onChange={action("onChange")}
     isLoading={boolean("isLoading", false)}
     hasError={boolean("hasError", false)}
+    multiple={boolean("has multiple files", true)}
```
pppemo: 🌷`Has multiple files?`, `Has error`, `Is loading?`, etc.. At least I've been doing it this way in other components :)

ugogo: Yep, good point.
Maybe we should have guidelines for that too 🤔

packages/react/src/components/FileUploader/index.js
ugogo: ✅

PR #46, opened 2019-07-31
Icons/better stories

ugogo:


packages/react/src/components/Icon/index.stories.js
```diff
@@ -2,25 +2,60 @@ import React from "react";
...
-    ))}
-  </div>
-));
+    </div>
+  );
+};
+
+const stories = storiesOf("Icons", module);
+
+stories.addDecorator(withKnobs);
```
pppemo: Hmm, so since we decided to add this decorator on global level. But apparently this change is in Checkbox component PR that has not yet been merged. Do you maybe wanna cherry pick this https://github.com/heetch/flamingo/pull/43/files#diff-2e6866a078d3aa8cca27fed58050c6cd and then remove this declaration in L56?
This PR won't need to get QAed so it will get merged faster

ugogo: That's a too big comment for me to cherry-pick. (That's why we should be careful with our commit history 🙊)
I'll just remove the `withKnobs` call here

PR #47, opened 2019-07-31
Implemented Radio component

pppemo:


packages/css/src/components/Radio.css
```diff
@@ -0,0 +1,67 @@
+.Radio {
+  position: relative;
+  appearance: none;
+  background: white;
+  width: 20px;
```
ugogo: `rem` values 😉

packages/react/src/components/Radio/index.js
```diff
@@ -0,0 +1,55 @@
...
+  children,
+  checked,
+  helper,
+  name,
+  value,
+  onChange,
+}) => {
+  const buildInputTag = () => (
+    <input
+      defaultChecked={checked}
```
ugogo: Is this a thing?
Shouldn't it be `checked={checked}`?

packages/css/src/components/Radio.css
pppemo: 👍

packages/react/src/components/Radio/index.js
pppemo: if you want the state of the `<input>` to be managed by the `<input>` itself, then you need to pass `defaultChecked` that would be initial value of the component.
And I guess this is what we need.

If however the state of the `<input>` would be managed outside of the `<input>` (that is in `state`) then you use `checked` and then react requires to pass `onChange` hook as well.

pppemo:
The disabled knob (and all other knobs in fact :P) were not working because local `withKnobs` was already deleted, but global was not yet set. I've added it locally, because - yeah - Juan might need it for testing. I guess once we're done with all the components we'll need to make sure we've deleted it everywhere locally and only the global declaration remained

packages/react/src/components/Radio/index.js
ugogo: 👍🏼

PR #48, opened 2019-07-31
Feature/component/toggle

pppemo:


packages/css/src/components/Toggle.css
```diff
@@ -0,0 +1,101 @@
+.Toggle {
+  position: relative;
+  width: 54px;
```
ugogo: Use `rem` values 🙏

packages/css/src/components/Toggle.css
```diff
@@ -0,0 +1,101 @@
...
+  transition-delay: 0.1s;
+}
+
+.Toggle.is-active .Toggle--bullet-label.is-off {
+  visibility: hidden;
+  transition: visibility 0.1s linear;
+}
+
+.Toggle .Toggle--bullet-label.is-on {
+  right: 18px;
```
ugogo: This could go into a variable as it's also used into `.Toggle--bullet-label.is-off`

packages/css/src/components/Toggle.css
```diff
@@ -0,0 +1,101 @@
...
+}
+
+.Toggle.is-active {
+  background-color: var(--color-rose);
+  border: 1px solid var(--color-rose);
+  transition: all 0.3s;
+}
+
+.Toggle.is-active .Toggle--bullet {
+  left: 100%;
```
ugogo: You can improve the animation by using the `translate` property.
Otherwise, every frame will cause a browser repaint

packages/react/src/components/Toggle/index.js
```diff
@@ -0,0 +1,62 @@
...
+  };
+
+  return (
+    <div className="ToggleContainer">
+      <div
+        className={cx("Toggle", { "is-active": isOn, "is-disabled": disabled })}
+        onClick={!disabled && handleToggle}
+      >
+        <div className="Toggle--bullet" />
+        <span className="Toggle--bullet-label is-on">{onLabel}</span>
```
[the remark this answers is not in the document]
ugogo: It will also avoid increasing the selector specificity when doing

```
.Toggle--bullet-label.is-off
```

packages/react/src/components/Toggle/index.js
```diff
@@ -0,0 +1,62 @@
...
+  const handleToggle = () => {
+    const newState = !isOn;
+    setIsOn(newState);
+    safeInvoke(onChange(newState));
+  };
+
+  return (
+    <div className="ToggleContainer">
+      <div
+        className={cx("Toggle", { "is-active": isOn, "is-disabled": disabled })}
```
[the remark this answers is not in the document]
pppemo: 👍

packages/react/src/components/Toggle/index.js
```diff
@@ -0,0 +1,62 @@
...
+        onClick={!disabled && handleToggle}
+      >
+        <div className="Toggle--bullet" />
```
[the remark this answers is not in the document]
pppemo: Yup! Dunno why

packages/react/src/components/Toggle/index.js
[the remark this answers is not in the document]
pppemo: If I do that change I'd need to duplicate a styling content of `.Toggle-bullet-label`

packages/react/src/components/Toggle/index.js
```diff
@@ -0,0 +1,62 @@
...
+        <span className="Toggle--bullet-label is-on">{onLabel}</span>
+        <span className="Toggle--bullet-label is-off">{offLabel}</span>
+      </div>
+      {(label || helper) && (
+        <div>
+          {label}
```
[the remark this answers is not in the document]
pppemo: Done!

packages/react/src/components/Toggle/index.stories.js
```diff
@@ -0,0 +1,44 @@
+import React from "react";
+import { storiesOf } from "@storybook/react";
+import { withKnobs, boolean, text } from "@storybook/addon-knobs";
+import { action } from "@storybook/addon-actions";
+import { withInfo } from "@storybook/addon-info";
```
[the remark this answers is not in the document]
pppemo: This particular one no, because `withInfo` would be added only to `Playground` stories

packages/react/src/components/Toggle/index.js
```diff
@@ -0,0 +1,62 @@
...
+          {helper && <p className="Toggle-helper">{helper}</p>}
+        </div>
+      )}
+    </div>
+  );
+};
+
+Toggle.propTypes = {
+  checked: PropTypes.bool,
+  onLabel: PropTypes.string,
```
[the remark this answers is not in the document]
pppemo: Dunno :) Need to check how that works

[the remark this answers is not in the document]
pppemo: I'm actually gonna remove this feature completely as per https://heetch.slack.com/archives/CE76359GQ/p1564572396009200 Juan decided to go with ON/OFF only for now, as it doesn't stretch anyways

packages/css/src/components/Toggle.css
pppemo: 👍

packages/css/src/components/Toggle.css
pppemo: What would the variable be?

packages/css/src/components/Toggle.css
pppemo: But `translate` only does the movement part, while there are other properties (colors mainly) that still need to be handled via `transform`

packages/react/src/components/Toggle/index.js
```diff
@@ -0,0 +1,52 @@
...
+  return (
+    <div
+      className={cx("ToggleContainer", {
+        "is-on": isOn,
+        "is-disabled": disabled,
+      })}
+    >
+      <div className="Toggle" onClick={!disabled && handleToggle}>
+        <div className="Toggle-bullet" />
+        <span className="Toggle-bullet-label is-on">ON</span>
```
ugogo:
```suggestion
        <span className="Toggle-bullet-label Toggle-bullet-label--on">ON</span>
```

The `is-` shouldn't be used here because it's not a state per se but rather a modifier 🙂 
It's no this item which is `is-on`, but the `ToggleContainer`

packages/react/src/components/Toggle/index.stories.js
```diff
@@ -0,0 +1,42 @@
...
+
+import Toggle from "./";
+
+storiesOf("Buttons/Toggle", module)
```
ugogo: Should this be in the `Buttons` stories or the `Form controls` one? 🤔

packages/react/src/components/Toggle/index.stories.js
```diff
@@ -0,0 +1,42 @@
...
+        label={text("Label", "Toggle label")}
+        helper={text("Label helper", "Label helper")}
+        onChange={state => action(`onChange - state: ${state}`)}
+      />
+    ))
+  )
+  .add("All states", () => (
+    <>
+      <div style={{ width: "100%" }}>
+        <Toggle label="Toggle" helper="Normal" />
```
ugogo: I'm having console errors for the `onChange` fn
https://www.dropbox.com/s/n3emevpjm9khzu2/Capture%20d%27%C3%A9cran%202019-08-12%2011.42.37.png?dl=0

packages/css/src/components/Toggle.css
```diff
@@ -0,0 +1,109 @@
...
+  border-radius: 10px;
+  cursor: pointer;
+  transition: all 0.3s;
+}
+
+.Toggle:hover {
+  border: 1px solid var(--color-granite);
+}
+
+.Toggle .Toggle-bullet {
```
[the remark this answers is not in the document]
pppemo: 👍

packages/css/src/components/Toggle.css
```diff
@@ -0,0 +1,109 @@
...
+  height: 1.25rem /* 20px */;
+  background-color: var(--color-white);
+  border: 1px solid var(--color-chatelle);
+  border-radius: 10px;
+  cursor: pointer;
+  transition: all 0.3s;
+}
+
+.Toggle:hover {
+  border: 1px solid var(--color-granite);
```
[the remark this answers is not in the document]
pppemo: Why would you remove border width param? It doesn't render line by default

packages/react/src/components/Toggle/index.js
pppemo: 👍

packages/react/src/components/Toggle/index.stories.js
pppemo: Dunno :/
No strong opinion. But in theory button would also be a part of form right?

packages/react/src/components/Toggle/index.stories.js
pppemo: Fixed

pppemo:
@ugogo would you mind giving a rough QA on this branch for ImageUploader and FileUploader? There some merging conflicts, and I want to make 200% nothing got broken

packages/css/src/components/Toggle.css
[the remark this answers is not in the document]
ugogo: It does 🙂 
https://github.com/heetch/flamingo/pull/48/files#diff-ad38b17b07f231626b7767a3a1ab0847R6

packages/css/src/components/Toggle.css
ugogo: Yeah I meant `transform`, not `translate` 🤦‍♂ 
Updating a `color` does not trigger any repaint so it's ok. But `left` doest.

packages/css/src/components/Toggle.css
```diff
@@ -0,0 +1,113 @@
...
+.ToggleContainer.is-on .Toggle-bullet {
+  left: 100%;
+  margin-left: -16px;
+  transition: all 0.3s;
+  background-color: var(--color-white);
+}
+
+.ToggleContainer.is-disabled .Toggle {
+  background-color: var(--color-grey);
+  border: 1px solid var(--color-grey);
```
ugogo: Same as above
```suggestion
  border-color: var(--color-grey);
```

packages/css/src/components/Toggle.css
```diff
@@ -0,0 +1,113 @@
...
+  display: flex;
+}
+
+.ToggleContainer > .Toggle {
+  margin-right: var(--space-l);
+}
+
+.Toggle-helper {
+  margin: 0;
+  padding: 0;
```
ugogo: Not needed

packages/css/src/components/Toggle.css
```diff
@@ -0,0 +1,113 @@
...
+  color: var(--color-granite);
+  font-size: var(--font-size--s);
+}
+
+.ToggleContainer .Toggle-labels {
```
ugogo:
```suggestion
.Toggle-labels {
```

packages/css/src/components/Toggle.css
```diff
@@ -0,0 +1,113 @@
...
+.ToggleContainer.is-disabled .Toggle-bullet {
+  background-color: var(--color-chatelle);
+}
+
+.ToggleContainer {
+  display: flex;
+}
+
+.ToggleContainer > .Toggle {
+  margin-right: var(--space-l);
```
ugogo: This should go with `.Toggle-labels`, so you don't have to worry about `Toggle` margin.
Adding labels will add margins. If there is no label, there is no margin 
And the specificity will be highly reduced ✨

packages/css/src/components/Toggle.css
ugogo: 🆙 , it will help scalability

packages/react/src/components/Toggle/index.js
```diff
@@ -0,0 +1,54 @@
...
+  };
+
+  return (
+    <div
+      className={cx("ToggleContainer", {
+        "is-on": isOn,
+        "is-disabled": disabled,
+      })}
+    >
+      <div className="Toggle" onClick={!disabled && handleToggle}>
```
ugogo: ⚠️ This is throwing console errors

packages/react/src/components/Toggle/index.js
```diff
@@ -0,0 +1,54 @@
...
+        <div className="Toggle-bullet" />
+        <span className="Toggle-bullet-label Toggle-bullet-label--on">ON</span>
+        <span className="Toggle-bullet-label Toggle-bullet-label--off">
+          OFF
+        </span>
+      </div>
+      {(label || helper) && (
+        <div className="Toggle-labels" onClick={!disabled && handleToggle}>
```
ugogo: ⚠️ This is throwing console errors

packages/react/src/components/Toggle/index.stories.js
ugogo: Dunno either, let's keep it there then. Or move it at the root, your call 🙂

packages/css/src/components/Toggle.css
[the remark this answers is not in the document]
pppemo: Ohhh sorry I didn't se you did this by `hover` i thought it's at `.Toggle` itself and that's why it didn't make any sense for me :)
changed!

packages/css/src/components/Toggle.css
pppemo: 👍

packages/css/src/components/Toggle.css
pppemo: 👍

packages/css/src/components/Toggle.css
pppemo: 👍

packages/css/src/components/Toggle.css
pppemo: I did it this way b/c even if there's no label I wanted to keep some horizontal distance between toggles if they were to be next to another. Even though it's doing it's job, I'm aware it's not very `all in place and order`. How about if I more this to the left of `Toggle-labels` and add small margin on left and right of `ToggleContainer`?

packages/react/src/components/Toggle/index.js
pppemo: 👍

packages/react/src/components/Toggle/index.js
pppemo: 👍

packages/css/src/components/Toggle.css
pppemo: 👍

packages/react/src/components/Toggle/index.stories.js
pppemo: moved at root

packages/css/src/components/Toggle.css
pppemo: 👍

pppemo:
All fixed. I only gave up on the transition/transform part as whatever I tried the combination of both didn;t want to work as expected :/

PR #52, opened 2019-08-12
Tweaked Select component to match latest specs

pppemo:


packages/react/src/components/Select/index.js
```diff
@@ -36,7 +36,11 @@ const Select = ({
         ))}
       </select>
 
-      <Icon icon="IconTopDown" className="FormEl-icon" size={ICON_SIZES.S} />
+      <Icon
+        icon="IconChevronDown"
```
ugogo:
```suggestion
        icon={Icon.IconChevronDown}
```

packages/react/src/components/Select/index.stories.js
```diff
@@ -34,9 +39,13 @@ storiesOf("Form controls/Select", module)
       </div>
     ))
   )
-  .add("With custom props", () => (
+  .add("Playground", () => (
     <>
```
ugogo: This `Fragment` is useless 😉

packages/react/src/components/Select/index.stories.js
```diff
@@ -34,9 +39,13 @@ storiesOf("Form controls/Select", module)
...
-      <Select {...defaultProps()} />
-      <Select {...defaultProps()} fill />
+      <Select
+        options={mapStringsToSelectOptions(
```
ugogo: 👌

packages/react/src/components/Select/index.js
pppemo: 👍

packages/react/src/components/Select/index.stories.js
pppemo: 👍 Leftover

ugogo:
Merged into #63

PR #53, opened 2019-08-21
Tooling improvements

ugogo:
- Create a unique command to launch every package
- Every package should handle its own linting and formatting tools


package.json
```diff
@@ -1,25 +1,19 @@
...
-      "pre-commit": "lint-staged"
-    }
-  },
-  "lint-staged": {
-    "*.{js,json,css,md}": [
-      "prettier --write",
-      "git add"
-    ]
+  "engines": {
+    "node": "^8.12.0 || >=9.7.0"
```
obedparla: I'm new here, but any reason we're using node 8 instead of newer like 12? 12 should be much faster than 8 and I don't see any dependencies that would make us stuck on 8.

packages/react/src/components/ListItem/index.js
```diff
@@ -24,7 +24,10 @@ const ListItem = ({
   onClick,
 }) => (
   <div
+    role="button"
+    tabIndex={0}
```
obedparla: a11y 🚀

package.json
ugogo: IIRC node v8.10 is the last supported version (like in CRA https://create-react-app.dev/docs/getting-started#creating-an-app), so it's more a matter of common agreement 😄 
And you can still use 12 because of the `>=9.7.0`

obedparla: Gotcha! thank you :)

Issue #57, opened 2019-08-26
Compass doesn't support usage of `vars` in :root


obedparla:
I couldn't find a specific bug filed for this since Compass-rails isn't being supported since 3 years ago, but after a lot of digging, I managed to get the CSS of Flamingo working on the admin.

The problem comes when trying to use a variable in `:root`, looks like Compass can't find its declaration at runtime or something.

```   
:root {
 --gradient-rr: linear-gradient(
            to right bottom,
            var(--color-rose),
            var(--color-razzmatazz)
    );
}
```
I needed to replace the variables for the real values and then it worked. It would look a bit ugly to have the same values twice, but unless we change Compass there's no avoiding it (I think).

I also found `--color-razzmatazz` and `--color-rose2` to be non-existent variables, they should be replaced or removed

obedparla:
https://github.com/heetch/flamingo/pull/58

PR #58, opened 2019-08-26
Fix Compass bug, unable to use variables in :root

obedparla:
I replaced two missing vars `--color-razzmatazz` and `--color-rose2` with `--color-rose` for this PR since I wasn't sure what they're supposed to be, but they can be anything you guys decide.

Issue https://github.com/heetch/flamingo/issues/57


obedparla:
Closed since a better fix was found by integrating Webpacker and not relying on Compass

PR #59, opened 2019-08-30
Add Table component

ugogo:


packages/react/src/components/Table/index.js
```diff
@@ -0,0 +1,158 @@
...
+                isSortedDesc,
+                render,
+              }) => (
+                <HeaderCell
+                  {...getHeaderProps(isSortable && getSortByToggleProps())}
+                  isSorted={isSorted}
+                >
+                  <Text>
+                    {render("Header")}
+                    {isSorted && isSortedDesc && <Icon icon="IconArrowUp" />}
```
pppemo:
```suggestion
                    {isSorted && <Icon icon={isSortedDesc ? "IconArrowUp" : "IconArrowDown"} />}
```

packages/react/src/components/Table/index.js
```diff
@@ -0,0 +1,158 @@
...
+                </HeaderCell>
+              )
+            )}
+          </HeaderGroup>
+        ))}
+      </thead>
+
+      <tbody>
+        {rows.map(row => (
+          <RowGroup {...prepareRow(row)} {...row.getRowProps()}>
```
pppemo:
```suggestion
          <RowGroup {...getRowProps(row)} {...row.getRowProps()}>
```

🌷
does this sound like a better name?

pppemo: b/c you have `getCellProps`

packages/react/src/components/Table/index.js
```diff
@@ -0,0 +1,158 @@
...
+};
+
+Table.defaultProps = {
+  className: undefined,
+  columns: [],
+  data: [],
+  isSortable: false,
+};
+
+Table.HeaderGroup = HeaderGroup;
```
pppemo: Ugo, just because this file grew this big, shouldn't each of these components be in a separate file,
and then this file would be doing 3 imports and then exporting everything as from one file?

packages/react/src/components/Table/index.js
ugogo: `prepareRow` is pulled from the main hook 😉 
```
const { headerGroups, rows, prepareRow } = useTable(
    {
      columns,
      data,
    },
    useSortBy
  );
```

packages/react/src/components/Table/index.js
ugogo: Create 3 files for less 10 lines each, IDK if it's worth it

packages/react/src/components/Table/index.js
ugogo: Yep 👍 
Done

packages/react/src/components/Table/index.js
pppemo: How about
```
const { headerGroups, rows, prepareRow: getRowProps } = useTable(
```
then?

Like, I'm not insisting, but for the consistency POV, could be nice.
Dunno, up to you to decide

packages/react/src/components/Table/index.js
pppemo: Saving couple of lines is one thing, but readability of the file itself could be of consequence.
As in, the navigation across this file is not as easy as it could be.
@obedparla @keanan what are you thoughts on this?

packages/react/src/components/Table/index.js
ugogo: `prepareRow` doesn't return any props, it actually "prepares" rows to be rendered 🙂

ugogo: I moved this part higher in the component so it's not confusing 👍

PR #62, opened 2019-09-13
Add Card component

ugogo:


packages/react/src/components/Card/index.js
```diff
@@ -0,0 +1,31 @@
...
+  <div
+    className={cx("Card", className, {
+      "is-selected": isSelected,
+      [`is-elevated--${elevation}`]: elevation,
+    })}
+    {...props}
+  />
+);
+
+Card.propTypes = {
```
obedparla: What about adding "Children" here?

packages/react/src/components/Card/index.js
```diff
@@ -0,0 +1,31 @@
...
+  isSelected: PropTypes.bool,
+};
+
+Card.defaultProps = {
+  className: undefined,
+  elevation: 0,
+  isSelected: false,
+};
+
+Card.ELEVATIONS = ELEVATIONS;
```
obedparla: Wouldn't it be cleaner to:

```
export {ELEVATIONS, Card}
export default Card 
```

Like this it makes me think I should/can do

```
import Card from ....
...
<>
   <Card elevation={Card.ELEVATIONS} />
<>
```

Which I'm not sure it's valid and can be confusing

packages/react/src/components/Card/index.js
```diff
@@ -0,0 +1,31 @@
...
+import PropTypes from "prop-types";
+import cx from "classnames";
+
+const ELEVATIONS = [0, 1, 2];
+
+const Card = ({ className, elevation, isSelected, ...props }) => (
+  <div
+    className={cx("Card", className, {
+      "is-selected": isSelected,
+      [`is-elevated--${elevation}`]: elevation,
```
pppemo:
```suggestion
      [`is-elevated--${elevation}`]: ELEVATIONS.includes(elevation),
```
?

packages/react/src/components/Card/index.js
pppemo: I think @obedparla what you want is actually achieved by the current implementation, and the suggestion would disable it

packages/react/src/components/Card/index.js
keanan: @pppemo won't that return true instead of the number?

keanan: ah sorry I see what's happening, you actually want it to return "true"

pppemo: Yeah, you want true. If it's a number, JS just casts it to boolean. But generally the value here is expected to be `boolean`.
And I thought that would be another guard (next to `propTypes`) that would cover a scenario, when user invokes `elevation={66609}` for example :)

keanan: Got you! Thanks for the insight.

ugogo: I was about to say that `0` is casts to `false`, but +1 for the extra scenario 🙂

packages/react/src/components/Card/index.js
obedparla: You guys are right 😄

PR #66, opened 2019-09-23
New component: TextType

ugogo:
Introducing the `UI` component.
The goal of this is component is to uniformize how the texts are displayed (like Juan is doing is the Typography table: https://www.notion.so/heetch/f65b272d3e5d48448519d0236043c91d?v=82d16164724a4618886166589079ff9d)


packages/react/src/components/ListItem/index.js
```diff
@@ -40,17 +43,29 @@ const ListItem = ({
...
         })}
       >
         {children}
-      </div>
-      {subtitle && <p className="ListItem-subtitle">{subtitle}</p>}
+      </UI>
+
+      {subtitle && (
+        <UI
+          type={LIST_ITEM_SIZES.MINI ? UI.TYPES.SubContent : UI.TYPES.Content}
```
pppemo: Shouldn't this condition be same as in L50?
It will always return true.

And if that's the case, shouldn't L50 and L60 become an arrow function?

packages/css/src/components/UI.css
```diff
@@ -0,0 +1,64 @@
+/* Headings */
+
+.UI--Heading1,
```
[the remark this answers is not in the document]
ugogo: Yeah, I kinda agree, what about `Typo`?
`Copy` looks more like something about i18n, and `String` is a reserved JavaScript keyword

packages/react/src/components/ListItem/index.js
ugogo: Oops yeah you're right

ugogo: Fixed! Alongside some proptypes

packages/css/src/components/UI.css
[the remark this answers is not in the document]
niven: If it's only text display, maybe Markup?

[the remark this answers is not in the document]
ugogo: `Markup` implies that its main function would be a markup cohesion when it's mainly about visual consistency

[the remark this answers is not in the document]
pppemo: `Typo` would impose that there's a mistake.
I like `Copy` or `Markup`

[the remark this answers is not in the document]
pppemo: And if none of above, how above `TextStyle`? `TextType`?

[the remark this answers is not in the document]
obedparla: Why not simply 'Text' ? I used to have a Text component previously and never had issues with misunderstanding its nature. If that's too broad then Pemo's suggestions are good I think

packages/css/src/components/Helper.css
```diff
@@ -1,5 +1,8 @@
 .Helper {
+  --text-color: var(--color-granite);
+
   display: block;
+  color: var(--text-color);
 }
```
obedparla: Where is `.Helper` used? I don't see why we'd need a class that adds color and display clock. Wouldn't it be better called `.BaseText` or something?

packages/css/src/components/Popover.css
```diff
@@ -0,0 +1,30 @@
+.Popover {
+  --spaceAround: var(--space-s);
+
+  opacity: 0;
+  transition: opacity 0.1s ease-out;
```
obedparla: Question: isn't `0.1s` too fast for an animation? I usually use .2s+.

packages/css/src/components/UI.css
```diff
@@ -0,0 +1,64 @@
...
+.UI--Heading2,
+.UI--Heading3,
+.UI--Heading4,
+.UI--Heading5,
+.UI--Heading6 {
```
obedparla: We have `Heading--x`, but also `UI--HeadingX`. Maybe `UI--Heading-X` for the sake of consistency? And/or changing the  `Heading--x` for  `Heading-x`. 

When using the React comps this doesn't matter much, but with the classes directly consistency is important IMO

packages/css/src/components/UI.css
```diff
@@ -0,0 +1,64 @@
...
+  font-size: var(--font-size--m);
+  line-height: var(--line-height--m);
+}
+
+.UI--Content,
+.UI--SubContent {
+  font-weight: var(--font-weight--normal);
+}
+
+.UI--ContentBold,
```
obedparla: Question: is there any specific reason we're using PascalCase rather than hyphens or something?

packages/react/src/components/Heading/README.md
```diff
@@ -13,9 +13,9 @@ export default () => <Heading>Content</Heading>;
 #### `as`: oneOf(`HEADING_LEVELS`)
 
 ```jsx
-import { Heading, HEADING_LEVELS } from "@heetch/flamingo-react";
+import { Heading } from "@heetch/flamingo-react";
 
-export default () => <Heading as={HEADING_LEVELS.h3} />;
+export default () => <Heading as={3} />;
```
obedparla: I REALLY LIKE THIS. Having to use constants all the time is a bit of a pain 😄 passing a number is 🚀

packages/react/src/components/Heading/index.js
```diff
@@ -2,24 +2,24 @@ import React from "react";
...
 import { HEADING_LEVELS } from "../../constants";
 
-const levels = Object.keys(HEADING_LEVELS);
-
-const Heading = ({ as: Component, className, ...props }) => (
-  <Component
-    className={cx("Heading", `Heading--${Component}`, className)}
+const Heading = ({ as, className, ...props }) => (
+  <UI
+    type={`Heading${as}`}
```
obedparla: Not for right now but dropping the thought here: something we can consider is not using the prop `as` to pass data, that's used by StyledComponents (and other libraries) and having it here conflicts with it (last time I tried it).

packages/react/src/components/Label/index.js
```diff
@@ -4,17 +4,26 @@ import React from "react";
...
 import cx from "classnames";
 
-const Label = ({ className, htmlFor, ...props }) => (
-  <label className={cx("Label", className)} htmlFor={htmlFor} {...props} />
+import { refShapes } from "../../constants";
+
+const Label = ({ className, forwardedRef, htmlFor, ...props }) => (
+  <label
+    className={cx("Label", className)}
+    htmlFor={htmlFor}
```
obedparla: Can `htmlFor={htmlFor}` be removed since `{...props}` would pass it?

packages/react/src/components/ListItem/index.js
```diff
@@ -22,77 +29,95 @@ const ListItem = ({
...
+  return (
     <div
-      className={cx("ListItem-valueContainer", {
-        "is-dark": valueType === LIST_ITEM_VALUES_TYPES.DARK,
-        "is-sub-dark": valueType === LIST_ITEM_VALUES_TYPES.SUB_DARK,
-        "is-error": valueType === LIST_ITEM_VALUES_TYPES.ERROR,
-        "is-success": valueType === LIST_ITEM_VALUES_TYPES.SUCCESS,
-        "is-strong-value": strongValue,
+      role="button"
+      tabIndex={0}
```
obedparla: so nice to see a11y

packages/react/src/components/Text/index.stories.js
```diff
@@ -11,7 +11,7 @@ storiesOf("Text", module)
       <Text>Default Text</Text>
 
       <strong>
-        <Text>Strong Text</Text>
+        <Text as="strong">Strong Text</Text>
```
obedparla: Confused what this does: why does it send `as="strong"` when is already inside a `<strong>` tag?

packages/react/src/components/UploaderImageItem/index.js
```diff
@@ -20,16 +21,23 @@ const UploaderImageItem = ({ file, overrides, ...props }) => {
   }, [file]);
 
   return (
-    <UploaderItem overrides={{ icon, ...overrides }} file={file} {...props} />
+    <UploaderItem
+      overrides={{ icon, ...overrides }}
+      file={file}
```
obedparla: Remove `file={file}` since `{...props}`?

packages/react/src/constants/index.js
```diff
@@ -65,3 +59,29 @@ export const LIST_ITEM_VALUES_TYPES = {
...
   SUCCESS: "success",
 };
+
+export const UI_TYPES = {
+  Heading1: "Heading1",
+  Heading2: "Heading2",
+  Heading3: "Heading3",
+  Heading4: "Heading4",
+  Heading5: "Heading5",
+  Heading6: "Heading6",
```
obedparla: Just a suggestion: would be nice to have these as `H<X>` rather than `Heading<X>`

obedparla:
Good work @ugogo! Really big PR with lots of improvements

packages/css/src/components/UI.css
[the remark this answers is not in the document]
ugogo: `Text` is actually a component on itself

packages/css/src/components/Helper.css
ugogo: `.Helper` is related to the `Helper` component, which needs some extra properties (like those ones)

packages/css/src/components/Popover.css
ugogo: I'll see that with Juan (designer in charge) during the final QA phases

packages/css/src/components/UI.css
ugogo: Yes, nice catch 👍 
Will use the `UI--Heading-X` pattern, even though I had to do some weird stuff on the UI component like
```
+    as={`h${as}`}
+    type={`Heading-${as}`}
```
😄

packages/css/src/components/UI.css
ugogo: Good question, I copied/pasted from Notion 🙈 
I'll camelCasify everything

packages/react/src/components/Heading/index.js
ugogo: Yeah, `as` should be only for HTML tag 🤔 
Let's iterate on this later

packages/react/src/components/Label/index.js
ugogo: Yes and no, because we want it `required` into the propTypes

packages/react/src/components/Text/index.stories.js
ugogo: 🤦‍♂ fixed

packages/react/src/components/UploaderImageItem/index.js
ugogo: Same as `Label`, required propTypes

packages/react/src/constants/index.js
ugogo: Since we switched to `Heading-<X>`, it could be weird to have `H-1` don't you think?

ugogo: Nevermind 🤦‍♂,
Done!

packages/css/src/components/UI.css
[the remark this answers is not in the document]
ugogo: `TextType` it is

PR #67, opened 2019-09-24
Component/popover

ugogo:


packages/react/src/components/Button/index.js
```diff
@@ -21,6 +29,10 @@ Button.propTypes = {
   as: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
   className: PropTypes.string,
   disabled: PropTypes.bool,
+  forwardedRef: PropTypes.oneOfType([
```
pppemo: In order to DRY, I guess this propType should be exported and imported similarly to how stubs are done

packages/react/src/components/Button/index.js
```diff
@@ -6,13 +6,21 @@ import { INTENTS } from "../../constants";
...
+  intent,
+  ...props
+}) => (
   <Component
     className={cx("Button", className, {
       "is-disabled": disabled,
       [`is-${intent}`]: intent,
     })}
     disabled={disabled}
+    ref={forwardedRef}
```
pppemo: Ohh, so we still will be needing to remember about that when implementing new components, huh?

pppemo: Is it really vital to do so since you're doing 
```
        React.cloneElement(children, {
          className: "Popover-wrapper",
          forwardedRef: ref,
```
?

Not saying it's not. Just trying to understand

packages/css/src/components/Popover.css
```diff
@@ -0,0 +1,30 @@
...
+  background-color: rgba(25, 1, 52, 0.8);
+  border-radius: 0.25rem /* 4px */;
+}
+
+.Popover-wrapper:hover + .Popover {
+  opacity: 1;
+}
+
+.Popover--top {
+  margin-bottom: var(--spacing);
```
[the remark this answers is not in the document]
ugogo: It defines the space between the content and the popover content

packages/react/src/components/Button/index.js
ugogo: Yeah we still need to do that because otherwise `forwardedRef` will end on the HTML element (via `{...props}`) and React will throw an error because this prop is not recognized as valid prop

packages/react/src/components/Button/index.js
ugogo: Done 👍🏼

packages/css/src/components/Popover.css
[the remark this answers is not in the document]
pppemo: Yeah, I'm just referring to the variable name `--spacing`
Shouldn't we have like `--spacing-m`, `--spacing-l` etc?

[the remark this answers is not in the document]
ugogo: `--spacing-m` and `--spacing-l` are global variables when `--spacing` is declared in `.Popover`

I'll rename it to `--spaceAround` to avoid confusion

ugogo:
@pppemo I agree that it could look weird and/or annoying, but using `ref` to access the DOM element is actually pretty common the React ecosystem. We can discuss that during the next weekly, but I'm afraid that there is no perfect solution

PR #68, opened 2019-10-03
Run prettier on everything

obedparla:
Update the `.prettierrc` config and run prettier with the new settings on all .js and .css files. The majority of the changes are from `" "` to `' '`.


obedparla:
> How come did we end up with a `'` instead of `"`?

We agreed on it on a FEW a few weeks back.

> And that in turn makes me wonder if all other settings are compatible with other projects
I don't see why it should cause an issue 🤔 changing `"` to `'` shouldn't cause issues. Prettier is very smart as well and when there is something like `"he said \"potato\""` it will fix that too without breaking anything

pppemo:
> We agreed on it on a FEW a few weeks back.
🤯

> I don't see why it should cause an issue
It causes no issue at all. Just thought it would make sense to have same config on every project. And I knew that we had `"` on every one of them. And apparently wasn't aware we've decided to change that.

Shall we then go all the way and get rid of `;`s as well? :)

PR #70, opened 2019-10-16
Release/v1

ugogo:


packages/react/src/components/Textarea/index.js
```diff
@@ -42,8 +42,11 @@ const Textarea = ({
         {...props}
       />
 
-      {isInvalid && <Icon icon='IconClose' {...defaultIconProps} />}
-      {isValid && <Icon icon='IconCheck' {...defaultIconProps} />}
+      {isInvalid && (
+        <Icon icon={Icon.ICONS.IconAlertOctagon} {...defaultIconProps} />
```
obedparla: Wouldn't it be better to pass down `className` directly? Since all `defaultIconProps` has is that 1 prop.

packages/css/src/components/Alert.css
```diff
@@ -0,0 +1,61 @@
...
+  );
+
+  position: relative;
+  margin: var(--space-m) 0;
+  padding: var(--space-xl) var(--iconContainerSize);
+  font-size: var(--font-size--m);
+  line-height: var(--line-height--m);
+  border-radius: var(--space-m);
+  background-color: var(--color-ui-element--primary);
+  border: 2px solid var(--mainColor);
```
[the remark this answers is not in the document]
ugogo: Exactly 🙂

packages/css/src/components/Button.css
```diff
@@ -44,78 +37,55 @@
...
-}
-
-.Button.is-primary:active,
-.Button.is-primary.is-active {
-  background-color: var(--color-ruby);
+.Button[disabled],
+.Button.is-disabled {
+  background-color: var(--color-ui-element-primary) !important;
+  color: var(--color-ui-text-tertiary) !important;
+  cursor: not-allowed !important;
```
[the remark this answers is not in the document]
ugogo: Those rules were overridden later because of the specificity.
But I added `!important` here because no matter what, we want disabled buttons to have this style

packages/react/src/components/FileUploader/index.js
```diff
@@ -2,36 +2,37 @@ import React from 'react';
...
-import { ICONS, INTENTS, refShapes } from '../../constants';
+import { refShapes } from '../../constants';
 import { safeInvoke } from '../../utils';
 
 const defaultTranslate = ({ defaultText }) => defaultText;
 
 const texts = {
   add_document: {
-    id: 'flamingo-file-uploader-add-document',
+    id: 'flamingo.file-uploader.add-document',
```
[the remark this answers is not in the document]
ugogo: Because with `-` you're forced to have every locale at the same level
```
{
  "flamingo-stuff-a": "foo",
  "flamingo-stuff-b": "bar",
  "app-related-heading": "baz"
}
```

Where with `.` you can nest locales
```
{
  "flamingo": {
    "stuff-a": "foo",
    "stuff-b": "bar",
  },
  "app-related-heading": "baz"
}
```

It adds a hierarchy layer, and you can fold/unfold the whole content into your editor

packages/react/src/components/Textarea/index.js
ugogo: Yep, done 👍

packages/react/src/components/Alert/index.js
```diff
@@ -25,12 +24,12 @@ const icons = {
 
 const types = Object.values(TYPES);
 
-const Alert = ({ title, children, forwardedRef, type, onClose }) => (
+const Alert = React.forwardRef(({ title, children, type, onClose }, ref) => (
```
obedparla: 👍  but also 😭 sorry

packages/react/.npmrc
```diff
@@ -0,0 +1 @@
+scripts-prepend-node-path=true
```
obedparla: Could you explain how this fixes the build?

ugogo: I basically followed the deployment logs 😄 
https://zeit.co/heetch-eng/flamingo/qz1yd047g

packages/react/src/components/Alert/index.js
ugogo: Don't be, that was a valid point!

groyoh:
Hey guys, I noticed a missing feature in the `Select`. There is no way to defined the `selected` value. It would be nice to have that before releasing v1 😄

ugogo:
👋 @groyoh 
You can do that by using the `value` prop

```
<Select value={x} .. />
```

groyoh:
Nevermind 🙈

packages/css/src/components/Button.css
```diff
@@ -131,3 +110,28 @@
...
 }
+
+/* Loading state */
+
+.Button-loadingState {
+  position: absolute;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
```
keanan: Hey man, why are we using `top`, `bottom`, `right` and `left` all together in places? can we not just use `flex-grow` and `flex-shrink` (shorthand: `flex: 0 0 100%`)? Just trying understand thanks

ugogo: Hey,
It's a tricky part because when a Button transitions to a `loading` state, we don't want the button to be resized. So the loading state has to **cover** the whole button, hence the absolute positions

PR #77, opened 2019-11-07
Feature/component/rich select

pppemo:
@ugogo I've only got the Readme pending, but it's now good for review


packages/css/src/components/Item.css
```diff
@@ -21,7 +21,7 @@
   cursor: pointer;
 }
 
-.f-Item.has-action:hover {
+.f-Item.has-action:hover, .f-Item.is-focused {
```
ugogo:
```suggestion
.f-Item.has-action:hover, .f-Item.has-action:focus {
```
?

packages/react/src/components/Dropdown/index.js
```diff
@@ -0,0 +1,130 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import ReactSelect from 'react-select';
+import Icon from '../Icon';
+import UiText from '../UiText';
+import { safeInvoke } from '../../utils';
+
+const customStyles = {
```
ugogo: Could you add those style via CSS instead of inline ones?

packages/react/src/components/Dropdown/index.js
```diff
@@ -0,0 +1,130 @@
...
+  control: () => ({
+    border: 0,
+    backgroundColor: 'transparent',
+    minHeight: 'initial',
+    display: 'flex',
+    justifyContent: 'space-between',
+  }),
+  indicatorSeparator: provided => ({
```
ugogo:
```suggestion
  indicatorSeparator: styles => ({
```

packages/react/src/components/Dropdown/index.js
```diff
@@ -0,0 +1,130 @@
...
+    ...provided,
+    display: 'none',
+  }),
+  indicatorContainer: provided => ({
+    ...provided,
+    marginRight: '1rem',
```
ugogo: Use :flamingo: vars?

packages/react/src/components/Dropdown/index.js
```diff
@@ -0,0 +1,130 @@
...
+    ref={innerRef}
+    {...innerProps}
+    icon={Icon.ICONS.IconChevronUpDown}
+    className='f-FormEl-icon'
+    size={Icon.SIZES.L}
+  />
+);
+
+const CustomSingleValue = ({ children }) => (
+  <UiText as='span' type={UiText.VARIANTS.content}>
```
ugogo: `Item`?

packages/react/src/components/Dropdown/index.js
```diff
@@ -0,0 +1,130 @@
...
+    {children}
+  </UiText>
+);
+
+const buildCustomOption = (onChange, isSelectable) => ({
+  data,
```
ugogo: `data: { labelComponent }` maybe?

packages/react/src/components/Dropdown/index.js
```diff
@@ -0,0 +1,130 @@
...
+    focused: isFocused,
+    onClick: e => {
+      safeInvoke(onChange, data.value);
+      safeInvoke(data.labelComponent.props.onClick, e);
+      isSelectable ? innerProps.onClick(e) : setValue(null);
+    },
+  },
+});
+
+const parseListItemObjectsIntoOptionObjects = objects =>
```
ugogo: I don't get what this fn is doing and it's quite verbose 😅

packages/react/src/components/Dropdown/index.js
```diff
@@ -0,0 +1,130 @@
...
+  [].concat(objects).map(object => ({
+    label: object.props.children,
+    value: object.props.optionValue || object.props.children,
+    labelComponent: object,
+  }));
+
+const Dropdown = React.forwardRef(
```
ugogo: 👍

packages/react/src/components/Dropdown/index.js
```diff
@@ -0,0 +1,130 @@
...
+  (
+    { className, onChange, children, placeholder, isSelectable, ...props },
+    ref,
+  ) => {
+    const [hasFocus, setHasFocus] = useState(false);
```
ugogo: Is this required?

packages/react/src/components/Dropdown/index.js
```diff
@@ -0,0 +1,130 @@
...
+const parseListItemObjectsIntoOptionObjects = objects =>
+  [].concat(objects).map(object => ({
+    label: object.props.children,
+    value: object.props.optionValue || object.props.children,
+    labelComponent: object,
+  }));
+
+const Dropdown = React.forwardRef(
+  (
+    { className, onChange, children, placeholder, isSelectable, ...props },
```
ugogo: Can an option not be selectable?

packages/react/src/components/Dropdown/index.js
```diff
@@ -0,0 +1,130 @@
...
+    ref,
+  ) => {
+    const [hasFocus, setHasFocus] = useState(false);
+
+    return (
+      <div className='f-FormEl-wrapper f-Select-wrapper'>
```
ugogo: `f-Select-wrapper` is not used anywhere
If you can `Select-wrapper` too in the `Select` component it'd be great, thanks!

packages/react/src/components/Dropdown/index.js
```diff
@@ -0,0 +1,130 @@
...
+        <ReactSelect
+          placeholder={placeholder}
+          options={parseListItemObjectsIntoOptionObjects(children)}
+          components={{
+            DropdownIndicator: CustomDropdownIndicator,
```
ugogo:
```suggestion
            DropdownIndicator,
```

packages/react/src/components/Dropdown/index.js
```diff
@@ -0,0 +1,130 @@
...
+            Option: buildCustomOption(onChange, isSelectable),
+            Placeholder: CustomSingleValue,
+            SingleValue: CustomSingleValue,
```
ugogo: Would it be possible to have
```
Placeholder,
SingleValue,
```
?

packages/react/src/components/Dropdown/index.js
```diff
@@ -0,0 +1,130 @@
...
+          }}
+          onFocus={() => setHasFocus(true)}
+          onBlur={() => setHasFocus(false)}
+          className={cx(
+            'f-FormEl',
+            'f-FormEl--withIcon',
+            'f-Select',
```
ugogo: I don't think `f-Select` is actually used somewhere

packages/react/src/components/Dropdown/index.js
```diff
@@ -0,0 +1,130 @@
...
+            'f-Dropdown',
+            className,
+            { 'is-focus': hasFocus },
+          )}
+          isSearchable={false}
```
ugogo: This should be moved into a prop

```suggestion
          isSearchable={isSearchable}
```

packages/react/src/components/Dropdown/index.js
```diff
@@ -0,0 +1,130 @@
...
+          styles={customStyles}
+          classNamePrefix='FlamingoDropdown'
```
ugogo:
```suggestion
          classNamePrefix='f-Dropdown-'
```

packages/react/src/components/Dropdown/index.js
```diff
@@ -0,0 +1,130 @@
...
+          ref={ref}
+          {...props}
+        />
+      </div>
+    );
+  },
+);
+
+Dropdown.displayName = 'Dropdown';
```
ugogo: 👌

packages/react/src/components/Dropdown/index.js
```diff
@@ -0,0 +1,130 @@
...
+
+Dropdown.propTypes = {
+  /** Make this true if you want this dropdown to behave like a select component */
+  isSelectable: PropTypes.bool,
+  className: PropTypes.string,
+  placeholder: PropTypes.string,
```
ugogo: I'm pretty sure you have some eslint errors with this, could you confirm?

packages/react/src/components/Dropdown/index.stories.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { withKnobs } from '@storybook/addon-knobs';
+import { action } from '@storybook/addon-actions';
+import { Dropdown } from '..';
+import { Item } from '..';
+
+const defaultProps = () => ({
```
ugogo: Yeah you don't need this, you can just add an `id` for each instance
And the action should exist in the `Playground` story

packages/react/src/components/Dropdown/index.stories.js
```diff
@@ -0,0 +1,44 @@
...
+  id: `dropdown-${Math.random()}`,
+  onChange: action('onChange'),
+});
+
+storiesOf('Dropdowns/Dropdown', module)
```
ugogo:
```suggestion
const stories = storiesOf('Dropdown', module)
```

Could you follow the same pattern as for every story?

- Use `const stories = `
- Add Heading
- Have a Playground story with as many knobs as possible as long as they make sense

packages/react/src/components/Dropdown/index.stories.js
```diff
@@ -0,0 +1,44 @@
...
+      <Item
+        type={Item.SIZES.NORMAL}
+        value='Take me there'
+        helper='Check what can really happend!'
+        onClick={action('onClick')}
+      >
+        Cars choice
+      </Item>
+      <Item
+        type={Item.SIZES.NORMAL}
```
ugogo: I'm pretty sure this is the default prop

packages/react/src/components/Item/index.js
```diff
@@ -35,7 +36,7 @@ const Item = React.forwardRef(
     const isMini = size === SIZES.MINI;
 
     const containerProps = onClick && {
-      onClick: safeInvoke(onClick),
+      onClick,
```
ugogo: Why did you remove the `safeInvoke` fn?

packages/css/src/components/Item.css
pppemo: Can't be, this is for the sake of `react-select` and it's executing focus on element by adding `is-focused`

packages/react/src/components/Dropdown/index.js
pppemo: I was considering it, but then it would require each attribute to be followed by `!important` to override original style. Do you think that would be better?

packages/react/src/components/Dropdown/index.js
pppemo: 👍

packages/react/src/components/Dropdown/index.js
pppemo: WDYM?

packages/react/src/components/Dropdown/index.js
pppemo: 👍

packages/react/src/components/Dropdown/index.js
pppemo: That's another problem since it's the inline style, flamingo vars are in CSS variables, so cannot be used here.
We could use flamingo vars, if these all were in CSS with `!important`.
Both solutions stink a bit. Dunno. What do you thinkk?

packages/react/src/components/Dropdown/index.js
pppemo: That is `UiText` because it represents the _selected_ value in the viewport of the select. Per design, even if the option itself is `Item` - meaning complex and rich, with arrows etc, only test is shown. This may be discussed with Juan further on if he thinks it's ok

packages/react/src/components/Dropdown/index.js
pppemo: Now that I've read it, yeah you're right, it's not that straightforward :)

So what it does:
1. What you're passing to `Dropdown` is an array of `Item`s
2. What you need to pass to `react-select` is a JSON label-value array.
3. So what it does it's wrapping every single `Item` into label-value object

Agree name is verbose. Can't think of anything better though. Have any ideas?

packages/react/src/components/Dropdown/index.js
pppemo: yes, b/c `react-select` doesn't have `:focus` selector itself. So if I want to change a style based on focus, I need to use `onFocus` and `onBlur`, store state, and apply `is-focused` style. See your first comment

packages/react/src/components/Dropdown/index.js
pppemo: That's for the case when the `Dropdown` will work as a menu (that was one of the requirements from Juan). Then once you click an option, I believe it shouldn't become "selected". You can click it, action will happen, but the value will not jump in the viewport of the component.

packages/react/src/components/Dropdown/index.js
pppemo: Done in both components!

packages/react/src/components/Dropdown/index.js
pppemo: 👍

packages/react/src/components/Dropdown/index.js
pppemo: I've changed `SingleValue`, but it also goes to `Placeholder` so L91 has to remain as is

packages/react/src/components/Dropdown/index.js
pppemo: Gone!

packages/react/src/components/Dropdown/index.js
pppemo: I'd leave it as is. Searching is not in the scope, and enabling this would require implementation of searching entire content of `Item` component in order to make it work intuitively.

packages/react/src/components/Dropdown/index.js
pppemo: That's actually not used anymore. Removed the line.

packages/react/src/components/Dropdown/index.js
pppemo: And you were pretty right! Made it `isRequired`

packages/react/src/components/Dropdown/index.stories.js
pppemo: 👍

packages/react/src/components/Dropdown/index.stories.js
pppemo: Done, yeah sorry didn't notice the patter.
I think it would be possible to implement a decorator to do it for us. Will add it to my TODO and take care of it when have time :)

packages/react/src/components/Dropdown/index.stories.js
pppemo: 👍

packages/react/src/components/Item/index.js
pppemo: I didn't have an understanding why, but this line was crashing when `Item` was used with `Dropdown` with `react-select` :/
However, existence of `onClick` is checked anyways in L37 right?

packages/react/src/components/Dropdown/index.js
ugogo: You should still be able to do `marginRight: 'var(...)',` 🤔

packages/react/src/components/Dropdown/index.js
ugogo: `getItems`?

packages/react/src/components/Dropdown/index.js
ugogo: 👍

packages/react/src/components/Dropdown/index.js
ugogo: 👍

packages/react/src/components/Item/index.js
ugogo: Yep good catch 👍

packages/react/src/components/Dropdown/index.stories.js
```diff
@@ -1,44 +1,100 @@
 import React from 'react';
 import { storiesOf } from '@storybook/react';
-import { withKnobs } from '@storybook/addon-knobs';
 import { action } from '@storybook/addon-actions';
 import { Dropdown } from '..';
 import { Item } from '..';
+import { boolean, select, text, number } from '@storybook/addon-knobs';
```
ugogo: Please have a look at your eslint errors...

packages/react/src/components/Dropdown/index.stories.js
```diff
@@ -1,44 +1,100 @@
 import React from 'react';
 import { storiesOf } from '@storybook/react';
-import { withKnobs } from '@storybook/addon-knobs';
 import { action } from '@storybook/addon-actions';
 import { Dropdown } from '..';
 import { Item } from '..';
```
ugogo: Would be nice to avoid importing from the `components/index` file as we might end with a circular dependencies
```
import Item from '../Item';
```

packages/css/src/components/Dropdown.css
```diff
@@ -0,0 +1,3 @@
+.f-Dropdown {
+  padding: 0 !important;
```
ugogo: Why `!important`?

packages/react/src/components/Dropdown/index.js
ugogo: Can it be achieved by increasing the CSS selector level?
`.f-Dropdown .foo.bar`?

packages/react/src/components/Dropdown/index.js
ugogo: 👍

packages/react/src/components/Dropdown/index.js
```diff
@@ -0,0 +1,128 @@
...
+  />
+);
+
+const SingleValue = ({ children }) => (
+  <UiText as='span' type={UiText.VARIANTS.content}>
+    {children}
+  </UiText>
+);
+
+const buildCustomOption = (onChange, isSelectable) => ({
```
ugogo:
```suggestion
const buildOpt = (onChange, isSelectable) => ({
```
?

PR #79, opened 2019-11-08
Use box-shadow instead of border for Buttons

ugogo:
This is needed because using a `border` property makes the button larger than expected (40px vs 38px)

### DIFF
--- a/packages/css/src/components/Button.css
+++ b/packages/css/src/components/Button.css
@@ -60,7 +60,7 @@
 .f-Button.is-outline {
   background-color: var(--textColor);
   color: var(--mainBgColor);
-  border: 1px solid var(--mainBgColor);
+  box-shadow: 0 0 0 1px var(--mainBgColor);
 }
 
 .f-Button.is-minimal {


obedparla:
I'm interested in the reasoning for this change @ugogo. I remember hating box-shadow as borders since it doesn't add space and can be annoying to debug + add correct spacing. Not opposed to it just want to hear your thoughts

ugogo:
👋 @obedparla 
The spacing is indeed the issue
When you have on the same line "normal" buttons and outline ones, they should have the same height so the layout does not mess up.

Normal button (38px): https://www.dropbox.com/s/tz8vou435ycq3fv/Screenshot%202019-11-12%2010.11.02.png?dl=0
Outline button (40px): https://www.dropbox.com/s/3u2ntuz6ysqh6n2/Screenshot%202019-11-12%2010.12.13.png?dl=0

We could indeed play with the padding, but we would end with something like 

```css
.f-Button {
  padding-top: calc(var(--f-space--m) - var(--border-size);
}
```

For a simpler solution, we decided (with Juan) to use box-shadows instead of borders.

Does it make sense to you? Any other-as-simple solution?

obedparla:
Got it! Thank you for the pics, they made it clear 😄.

Apart from the `calc(var(--f-space--m) - var(--border-size);` you mentioned I don't know of any better. So box-shadow sounds like a good one!

PR #81, opened 2019-11-14
Component/grid

ugogo:
This one is a tricky one, if you are reviewing it please play with the stories to see if everything's working as expected 🙏


ugogo:
@pppemo Looks like you've misread the PR. The `ColInner` component is a purely decorative one used only in the stories.

PR #83, opened 2019-11-18
Added possibility to show an icon on Input right handside

pppemo:
This is per design https://www.figma.com/file/YOIkND79so4MHXbSt43kBzPY/New-Login-Method?node-id=323%3A713

when a lock icon is required to be presented on password input

### DIFF
--- a/packages/react/src/components/Input/index.js
+++ b/packages/react/src/components/Input/index.js
@@ -42,7 +42,7 @@ const Input = React.forwardRef(
           variant={UiText.VARIANTS.content}
           className={cx('f-FormEl', 'Input', className, {
             ...classes,
-            'f-FormEl--withIcon': isValid || isInvalid || icon,
+            'f-FormEl--withIcon': !!icon,
           })}
           disabled={isDisabled}
           id={id}
--- a/packages/react/src/components/Input/index.js
+++ b/packages/react/src/components/Input/index.js
@@ -17,7 +17,7 @@ const Input = React.forwardRef(
       onChange,
       placeholder,
       valid: isValid,
-      inputIcon,
+      icon,
       ...props
     },
     ref,
@@ -54,8 +54,8 @@ const Input = React.forwardRef(
           <Icon icon={Icon.ICONS.IconCheck} className='f-FormEl-icon' />
         )}
 
-        {!isValid && !isInvalid && inputIcon && (
-          <Icon icon={inputIcon} className='f-FormEl-icon' />
+        {!isValid && !isInvalid && icon && (
+          <Icon icon={icon} className='f-FormEl-icon' />
         )}
       </div>
     );
@@ -67,7 +67,7 @@ Input.displayName = 'Input';
 Input.propTypes = {
   className: PropTypes.string,
   disabled: PropTypes.bool,
-  inputIcon: PropTypes.oneOf(icons),
+  icon: PropTypes.oneOf(icons),
   id: PropTypes.string.isRequired,
   invalid: PropTypes.bool,
   onChange: PropTypes.func.isRequired,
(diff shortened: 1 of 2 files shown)


packages/react/src/components/Input/index.stories.js
```diff
@@ -51,6 +54,7 @@ stories.add('Playground', () => (
     invalid={boolean('Is invalid?', false)}
     valid={boolean('Is valid?', false)}
     disabled={boolean('Is disabled?', false)}
+    inputIcon={select('Input icon', ICONS, ICONS.None)}
```
ugogo: I'd go with `icon` 
```suggestion
    icon={select('Icon', ICONS, ICONS.None)}
```
Because `<Input inputIcon="x" />` feels a bit repetitive
Where `<Input icon="x" feels more natural IMO

pppemo: Dang that was my first choice!!! :-D
Changed!

packages/react/src/components/Input/index.js
```diff
@@ -50,6 +53,10 @@ const Input = React.forwardRef(
         {isValid && (
           <Icon icon={Icon.ICONS.IconCheck} className='f-FormEl-icon' />
         )}
+
+        {!isValid && !isInvalid && icon && (
```
ugogo: Does this actually work?
Shouldn't it be 
```suggestion
        {!isValid || !isInvalid || icon || (
```

ugogo: Nevermind 🤦‍♂ 
Should we export a function to get only 1 icon and clear this code a bit?
Like
```
const getIcon = () => {
  if (isInvalid) return '...'
  if (isValid) return '...'
  return icon
}
```
WDYT?

obedparla: I don't get this code, am I missing something? `!isValid && !isInvalid`, isn't that the same as a single `!isValid`?

obedparla: 🤦‍♂ ok, now I see they're two different vars. Maybe consider changing the name? Two people got facepalmed with this haha. I agree with Ugo's proposal

packages/react/src/components/Input/index.js
```diff
@@ -25,6 +28,14 @@ const Input = React.forwardRef(
       'is-invalid': isInvalid,
     };
 
+    const getInputIcon = () => {
```
ugogo: You don't have to prefix everything, you already are in an Input context 😉 

```suggestion
    const getIcon = () => {
```

packages/react/src/components/Input/index.js
```diff
@@ -25,6 +28,14 @@ const Input = React.forwardRef(
...
+      if (isValid) return Icon.ICONS.IconCheck;
+      if (isInvalid) return Icon.ICONS.IconAlertOctagon;
+      return icon;
+    };
+
+    const inputIcon = getInputIcon();
```
ugogo:
```suggestion
    const icon = getIcon();
```

packages/react/src/components/Input/index.js
```diff
@@ -43,13 +54,7 @@ const Input = React.forwardRef(
...
         />
 
-        {isInvalid && (
-          <Icon icon={Icon.ICONS.IconAlertOctagon} className='f-FormEl-icon' />
-        )}
-
-        {isValid && (
-          <Icon icon={Icon.ICONS.IconCheck} className='f-FormEl-icon' />
-        )}
+        {inputIcon && <Icon icon={inputIcon} className='f-FormEl-icon' />}
```
ugogo:
```suggestion
        {icon && <Icon icon={icon} className='f-FormEl-icon' />}
```

PR #85, opened 2019-11-21
Component/link

ugogo:
- Add `Link` component
- Update colors
- Upgrade dependencies


packages/css/src/shared/vars.css
```diff
@@ -11,13 +11,14 @@
   --f-color-element--primary: #fff;
   --f-color-element--secondary: #685879;
   --f-color-element--tertiary: #b3abbc;
-  --f-color-element--inactive: #f2f2f2;
+  --f-color-element--inactive: #eaeaea;
```
pppemo: Just checking, designers aware of that? :)

ugogo: Of course, it's a request from Juan actually 😉

PR #87, opened 2019-11-26
Fix Item onClick prop triggering

ugogo:


packages/react/src/components/Item/index.js
```diff
@@ -35,7 +35,7 @@ const Item = React.forwardRef(
     const isMini = size === SIZES.MINI;
 
     const containerProps = onClick && {
-      onClick: safeInvoke(onClick),
+      onClick: () => safeInvoke(onClick),
```
obedparla: What about doing the anonymous function in safeInvoke itself? Easier to use throughout the codebase:

`return isFunction ? () => fn(...args) : null;`

Rather than when calling safeInvoke

obedparla: Don't we want the app to break if we send something other than a function to an `onClick` prop though? I mean React itself would break, and I would expect it to break rather than to fail silently, sounds like a like to create a hard to debug issue, whereas if it explodes it's obvious

ugogo: So are you suggesting that we get rid of the `safeInvoke` util?

pppemo: Dang, I remember `Dropdown` was failing when `onClick: () => safeInvoke(onClick)`
Why do you need to revert it?

ugogo: @pppemo Because the `onClick` prop was fired immediately after the rendering

pppemo: Yeah, saw the thread on Slack. Ok that makes sense

obedparla: Yeah I suggest we remove the util altogether. We should stick to the default behavior on these things, especially since onClick props are made by the dev and any error will be caught during development

keanan: Nice one guys, agreed.

ugogo: Works for you @pppemo?

ugogo: Agreed on Slack ✅

PR #89, opened 2019-11-29
Add Overlay and SidePanel components

ugogo:


packages/react/src/components/SidePanel/index.js
```diff
@@ -0,0 +1,81 @@
...
+const SidePanel = ({
+  children,
+  className,
+  footer,
+  isOpen: isOpenProp,
+  onClose,
+  title,
+  ...props
+}) => {
+  const [isOpen, setIsOpen] = React.useState(isOpenProp);
```
pppemo:
```suggestion
  const [isOpened, setIsOpened] = React.useState(isOpenProp);
```

packages/react/src/components/SidePanel/index.js
```diff
@@ -0,0 +1,81 @@
...
+
+  const toggleIsOpen = () => setIsOpen(!isOpen);
```
pppemo:
```suggestion
  const toggleIsOpened = () => setIsOpen(!isOpen);
```

packages/react/src/components/SidePanel/index.js
```diff
@@ -0,0 +1,81 @@
...
+
+  const handleClose = () => {
+    toggleIsOpen();
+    onClose();
```
pppemo: What if `onClose === undefined`?

packages/react/src/components/SidePanel/index.js
```diff
@@ -0,0 +1,81 @@
...
+    setIsOpen(isOpenProp);
+  }, [isOpenProp]);
+
+  return (
+    <>
+      <Overlay
+        className={cx({
+          'is-hidden': !isOpen,
+        })}
+        onClick={handleClose}
```
pppemo: 🌷 How about a prop `overlayClickClosesPanel` set default to `true`?

packages/react/src/components/SidePanel/index.js
```diff
@@ -0,0 +1,81 @@
...
+      />
+
+      <div
+        className={cx('f-SidePanel', className, {
+          'is-open': isOpen,
```
pppemo:
```suggestion
          'is-openen': isOpened,
```

packages/react/src/components/SidePanel/index.js
```diff
@@ -0,0 +1,81 @@
...
+        })}
+        {...props}
+      >
+        <div className='f-SidePanel-header'>
+          <Heading level={4} as='h1'>
```
pppemo: Ok, I'm getting confused by our text classes again.
Heading level 4 - I read as `h4`
But that's followed by `h1`.
Either it's a mistake in this line or our `Heading` class is messed up and we need to fix it

packages/react/src/components/SidePanel/index.js
pppemo: Oh, just saw the defaults. Yeah that's nifty.
Although it builds an useless method in memory :P

packages/react/src/components/SidePanel/index.js
```diff
@@ -0,0 +1,81 @@
...
+  );
+};
+
+SidePanel.propTypes = {
+  children: PropTypes.node,
+  className: PropTypes.string,
+  footer: PropTypes.node,
+  isOpen: PropTypes.bool,
+  onClose: PropTypes.func,
+  title: PropTypes.node,
```
pppemo: Can `title` be more than text?

packages/react/src/components/SidePanel/index.js
ugogo: I'd rather go with `should` or `close` at the start because if we want to add other conditions (like close the panel when pressing ESC), it should have the same pattern.
So what about
`shouldCloseOnOverlayClick`
`closesWithOverlay`
`closesOnOverlayClick`
?

packages/react/src/components/SidePanel/index.js
ugogo: `as` is what component should be rendered
`level` is for styling purposes.

Since we don't want to style a component by its tag, we are relying on the `level` prop

By default `<Heading level={4} />` will render a `h4` with styles of a `h4`
But if you want this component to be a `h1` while keeping the same styles, you'll use the `as` prop

packages/react/src/components/SidePanel/index.js
ugogo: Yep, and that's why `node` is permissive enough in this case

packages/react/src/components/SidePanel/index.js
ugogo: Yeah it's fine 🙂

packages/react/src/components/SidePanel/index.js
pppemo: I like `closesOnOverlayClick`

packages/react/src/components/SidePanel/index.js
pppemo: 👍

packages/react/src/components/SidePanel/index.js
pppemo: 👍

packages/react/src/components/SidePanel/index.js
ugogo: ✅

ugogo:
@obedparla I knew someone would raise this issue, that's why it's in a CSS var, so you can update it into your project 🙂 
https://github.com/heetch/flamingo/pull/89/files#diff-788a7f3d0fc96b7a204aa033ce50443aR77

PR #90, opened 2019-12-03
Fix unprefixed classes

ugogo:


ugogo:
@obedparla I removed them because
- they weren't use
- they don't follow Flamingo CSS classes naming pattern

PR #92, opened 2019-12-18
Tweaked styles

pppemo:


ugogo:
@pppemo Could you provide more context? Why do you need this?

pppemo:
Because if this component is placed in another component that has:
```
text-align: center

f-Button: {
width: 100%
}
```

this is what happens
image

packages/css/src/components/Alert.css
```diff
@@ -11,6 +11,7 @@
 }
 
 .f-Alert {
+  text-align: left;
```
ugogo: Move this line after the variables

pppemo: done

PR #95, opened 2020-01-13
Form fields/imps

ugogo:


packages/react/src/components/InputField/index.js
```diff
@@ -1,13 +1,31 @@
...
 
-const InputField = React.forwardRef((props, ref) => (
-  <FormField {...props} ref={ref}>
-    <Input {...props} />
-  </FormField>
-));
+const InputField = React.forwardRef(
+  ({ label, helper, inputRef, ...props }, ref) => (
+    <FormField label={label} helper={helper} {...props} ref={ref}>
+      <Input ref={inputRef} {...props} />
```
keanan: If we sending props to both `FormField` and `Input` will they not get the same required `id`?

ugogo: That's the point 🙂 
The `FormField` needs this id for the `Label`
https://github.com/heetch/flamingo/blob/master/packages/react/src/components/FormField/index.js#L11

Whereas the `Input` will use it as a regular id `id={id}`

keanan: Cool got you, I see `FormField` is using it for the label, thanks

PR #96, opened 2020-01-13
Fix Alert title prop

ugogo:


packages/react/src/components/Alert/index.js
```diff
@@ -60,15 +60,16 @@ Alert.displayName = 'Alert';
 
 Alert.propTypes = {
   children: PropTypes.node,
-  type: PropTypes.oneOf(types),
-  title: PropTypes.string.isRequired,
   /** Method that is triggered when X button on the alert is clicked. Most likely to be used for hiding the alert. If not provided, no X button is gonna be shown. */
   onClose: PropTypes.func,
+  title: PropTypes.node,
```
keanan: Would it be better to have oneOf(string, number) instead of `node` for title?

ugogo: Yeah, but what if you need to use a React component for that?
Like an `I18n` component for example
`node` covers every case (strings, number, React component, etc)

keanan: We are not using the i18n component anymore just sending down props, but I see what you say, doesn't really make a huge difference.

PR #97, opened 2020-01-15
Dependencies/upgrade

ugogo:


ugogo:
On hold, needs to use `styled-components`

ugogo:
Done into https://github.com/heetch/flamingo/pull/138

PR #98, opened 2020-01-17
[FRONT-93] Add Selector component

tamarasaurus:
This PR adds the `Selector` component

selector

Spec: https://www.notion.so/heetch/Selector-3bb2d672a4f647fbaf9355bd5480a48a
Figma: https://www.figma.com/file/VOup0kJcIWzAOGA0rh62yI/%F0%9F%A6%84-Flamingo-Web?node-id=967%3A1417


package-lock.json
```diff
@@ -0,0 +1,6550 @@
+{
```
ugogo: Should we remove this file since we already have a `yarn.lock` file?

packages/css/src/shared/vars.css
```diff
@@ -1,6 +1,7 @@
 :root {
   --f-color-brandPrimary: #ff2773;
   --f-color-brandPrimary--light: #ff4d95;
+  --f-color-brandPrimary--inactive: #fc99bc;
```
ugogo: 👌

packages/react/src/components/Selector/index.js
```diff
@@ -0,0 +1,97 @@
...
+
+      setActiveIndex(index);
+      onChange(option.value);
+    };
+
+    return (
+      <div className='f-FormEl-wrapper'>
+        <UiText
+          as='div'
+          variant={UiText.VARIANTS.content}
```
ugogo: Not needed as it's the default prop
https://github.com/heetch/flamingo/blob/master/packages/react/src/components/UiText/index.js#L41

packages/react/src/components/Selector/index.js
```diff
@@ -0,0 +1,97 @@
...
+      value: PropTypes.oneOfType([PropTypes.string, PropTypes.number])
+        .isRequired,
+    }),
+  ),
+};
+
+Selector.defaultProps = {
+  className: undefined,
+  disabled: false,
+  defaultIndex: null,
```
ugogo:
```suggestion
  defaultIndex: 0,
```

Do we want to have the first item selected by default?

packages/react/src/components/Selector/index.js
```diff
@@ -0,0 +1,97 @@
...
+          id={id}
+          name={id}
+          onChange={onChange}
+          ref={ref}
+          {...props}
+        >
+          {options.map((option, index) => (
+            <div
+              key={option.value}
+              role='button'
```
ugogo: Why not using a `<Button />` here?
But using one, you could also inherit from all its styles (active, disabled, etc)

packages/react/src/components/Selector/index.js
tamarasaurus: Nope, apparently not, it's optional

package-lock.json
tamarasaurus: Oh yeah, my bad :)

packages/react/src/components/Selector/index.js
tamarasaurus: You're right, I can try that instead

tamarasaurus: So I tried using `<Button />`. I think it doesn't help much because I still need to override the styles of the states and add a couple of missing ones. I'll leave it this way

obedparla:
Hey @tamarasaurus , a few edge cases.

- When the string is too long, it touches the edges and looks ugly. 
- Same happens when the string is just long enough
- I guess the `...` is on purpose but I'd challenge it, if it's an option to users, how are they supposed to know what comes after the `...`?

obedparla:
image

tamarasaurus:
@obedparla According to the spec, short labels should be part of the design, otherwise the `Select` component is used. 

> You use it when the options have short labels and it helps the user to make a faster decision to see all the options.
> If the options have long labels, you can use the Select component.
> 

Is it up to us to support these edge cases if they don't conform to the spec?

obedparla:
I guess the decision is up to design, but feels buggy for me to have such a brittle component. If it solves the issue let's 👍 and modify it later if we have to

tamarasaurus:
@obedparla I fixed it anyway, by adding the padding to keep the alignment :+1: Just curious about the process :)

obedparla:
Yeah I think it's up to us to support these edge cases, as long as the component fit within the design :) (like using small labels as you said)

packages/css/src/components/Selector.css
```diff
@@ -0,0 +1,76 @@
+.f-Selector {
+  display: flex;
+  flex-direction: row;
+  width: 288px;
```
ugogo: Don't we want this to be flexible?

ugogo: And maybe use `rem` values 😬

PR #100, opened 2020-01-30
Update Selector component with "valid" and "invalid" states

obedparla:
The Selector component didn't have `valid` and `invalid` states but I need them for HMTD


tamarasaurus:
Looking good, and thank for adding the rest of the states to the `SelectorField` storybook page :clap:

PR #102, opened 2020-02-12
Fix engines in package.json

ugogo:
Set node v12.14.0 as the minimal required version, but also allow greater ones (13.x at the moment)


ugogo:
Not Heetch CI, but I tried to used Flamingo on another computer and its Node version was 13.something, so I wasn't able to run `yarn install` into my project ☹️

PR #103, opened 2020-02-14
Add new icons

ugogo:
- `IconArrowDownCircle`
- `IconArrowLeftCircle`
- `IconArrowRightCircle`
- `IconArrowUpCircle`
- `IconEyeClose`
- `IconEyeOpen`


ugogo:
@obedparla 👋 
The React components are created from the SVG files via the `icons` script

https://github.com/heetch/flamingo/blob/master/packages/react/package.json#L25
https://github.com/heetch/flamingo/blob/master/packages/react/scripts/icons.js

Thanks for the review!
There is one icon which is not rendering as expected ☹️ 
Screenshot 2020-02-17 09 49 53

PR #110, opened 2020-03-03
Migration to Styled Components base changes

obedparla:
## Important code changes:

- All code now lives within `src` and there are only React components now, rather than `css` and `react` repos.
- There's no separate `CSS` anymore. The only CSS we have is for exporting flamingo's CSS variables and is exported alongside the React Components. 
- Moved and modified relevant files for development such as `babel.rc`, `storybook` config and so on to the root folder.

If you want to see the changes, find the components I created. Lots of changes due to the modification of folder structure.

## Completed:

I aimed at doing all the base components that are used by other components so that no one has to step on each other toes. There should be an example for every use case already implemented.

### Completed components:

- Button
- FormElement -> base component used for form fields (input, text area, etc) that used to be a class.
- FormElementWrapper -> wrapper component used for form fields that used to be a class. 
- FormField -> base component
- Heading
- Helper
- Icon
- Input. Check this for guidance on how to implement a form element. (the idea is to pass props rather than rely on classes for highlighting and such)
- InputField
- Label
- Spinner
- UiText


deprecated/components/FileUploaderField/index.stories.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { text } from '@storybook/addon-knobs';
+import { action } from '@storybook/addon-actions';
```
OctaveStoufflet: What is a .stories file please ? :)

obedparla:
@heetch/frontend squad, let's discuss something important:

There are two options for migrating:

### Leave the structure
We leave the folder structure as it is in the current repo, so we'd still have the `css` and `react` folders. With the `css` folder not changing at all and the `react` component becoming Styled Components

**Pros**
- The project structure doesn't change that much
- We keep the `CSS` files intact, with the possibility of modifying them if we need to
- No need to modify deployment settings since it already works with two folders

**Cons**
- This is a new version of Flamingo, and we'd leave a "deprecated" package (the CSS) within `master` alongside the new version (styled-components)
- As a developer, I'd find it strange to see an OSS library hosting a deprecated version of code alongside the new version

### Modify the structure
Modify everything and move the `react` components to the root, removing `CSS`. (this is the route I took for this PR)

**Pros**
- We need the repo clean by only having the latest version of the packages in `master`. The old version can be behind a branch/release tag to easily access it
- Better DX (IMO) since there's only 1 source of truth, the component itself, which should contain everything
- This is the approach OSS libraries take when doing major version changes, and we need to remember Flamingo is an OSS library

**Cons**
- The structure changes dramatically
- Harder to review (but IMO we should review this in the IDE and check the new components, rather than in Github)
- We'll need to change the deployment settings
- We won't be modifying the CSS files anymore

### Opinion

What do you think?  I didn't consider the first option until Ugo pointed it out, and it's a valid point.

I prefer the second option (changing everything) since this is a major version of the design system with a complete rewrite of its core, which for me means modifying it in a way to match the expectations.

A question that we didn't discuss so far: are we going to use the `CSS` files anymore? If yes, we'll need to keep them up to date or modify them, that would be a strong pro for the first option. I don't think we should do this since the point of moving to `SC` was to stop using `CSS` files directly.

ugogo:
👋 @obedparla 
I have mix feelings about the whole project reorganization
On one hand, if you are currently using the CSS package, you might need to navigate across those files.
On the other hand, I agree that the organization should better represent what we are serving

Since GitHub is keeping a history of everything, I think we can update the README with:
- A warning explaining why the CSS package is deprecated
- A link to a previous tag so anyone can still navigate through those files

WDYT?

NB: I still think this should be done in different PRs 😬 

PR1: Update the README, remove the CSS package
PR2: Start updating components (no need to move all of them into a deprecated folder)

obedparla:
> A warning explaining why the CSS package is deprecated
A link to a previous tag so anyone can still navigate through those files

I love this approach. Let's do it this way

> PR1: Update the README, remove the CSS package

Agreed

> PR2: Start updating components (no need to move all of them into a deprecated folder)

I moved them in order to make it easy to know which component had already been implemented and which wasn't, but mainly because we still need to be able to access the `css` files in order to migrate them and to compare with the old component without overriding it. 

If I remove them completely you'd need to have 2 flamingo projects open (or GitHub, or something) to access the CSS (which was removed) 😅. With all of them in the same folder, it's hard to differentiate. We will ofc remove the whole "deprecated" folder once we finish the migration. 

It's basically to facilitate DX in the migration. WDYT @ugogo 

I'll modify and open the new PRs once we agree on the last point ^

ugogo:
Works for me @obedparla 🙂 
If you could just wrap the whole components-moving-into-a-deprecated-folder into a single commit, that would be perfect!
Thanks for taking the lead on this btw!

deprecated/components/FileUploaderField/index.stories.js
keanan: We use storybook to isolate and showcase Flamingo components https://storybook.js.org/docs/basics/introduction/

src/components/Button/index.js
```diff
@@ -0,0 +1,183 @@
+import React from 'react';
```
tamarasaurus: Is there any reason not to use `.jsx` as a file extension? I don't have a strong preference just wondering

src/components/Button/index.js
```diff
@@ -0,0 +1,183 @@
...
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import cx from 'classnames';
```
tamarasaurus: Is that for backwards compatibility?

src/components/Button/index.js
```diff
@@ -0,0 +1,183 @@
...
+  MINIMAL: 'minimal',
+};
+
+const intents = Object.values(INTENTS);
+const variants = Object.values(VARIANTS);
+
+const styles = {
+  color({ disabled, variant, intent }) {
+    if (disabled) {
+      return 'var(--f-color-text--tertiary) !important';
```
tamarasaurus: Do we plan to have a theme in JS at some point?

src/components/Spinner/index.stories.js
```diff
@@ -0,0 +1,17 @@
...
+import Icon from '../Icon';
+import Spinner from '.';
+
+const sizes = Object.values(Icon.SIZES);
+const stories = storiesOf('Spinner', module);
+
+stories.add('Playground', () => (
+  <>
+    <Heading>Spinner</Heading>
+    <Spinner size={select('Size', sizes, Icon.SIZES.M)} />
```
tamarasaurus: Even though `size` is passed to `Icon` should we be more explicit and put it in prop types?

src/components/UiText/index.js
```diff
@@ -0,0 +1,129 @@
...
+  contentBold: 'contentBold',
+  subContent: 'subContent',
+  subContentBold: 'subContentBold',
+};
+
+const variants = Object.keys(VARIANTS);
+
+const styles = {
+  fontSize({ variant }) {
+    if (variant === VARIANTS.h1) {
```
tamarasaurus: :tulip:  Do you wanna enforce this style of conditionals? I find objects to be way more concise

```javascript
const variants = {
[VARIANTS.h1]: 'var(--f-fontSize--xxl)',
[VARIANTS.h2]: 'var(--f-fontSize--xl)'
.....
}

return variants[variant] || undefined 
```
What do you think?

src/components/Icon/index.js
```diff
@@ -0,0 +1,66 @@
...
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import styled from 'styled-components';
+
+import * as SVGS from '../../constants/icons';
+
+const NAMES = Object.keys(SVGS);
+
+const ICONS = NAMES.reduce(
+  (icons, icon) => ({
```
tamarasaurus: Nice for when we add new icons :heavy_plus_sign:

.travis.yml
```diff
@@ -0,0 +1,4 @@
+language: node_js
+node_js:
+  - 9
```
tamarasaurus: Doesn't flamingo run on node 12.14.1 ?

src/components/Button/index.js
obedparla: The reason is that rollup is setup with `js` files. I  tried briefly to move them into JSX but didn't manage (I don't know much about rollup) and didn't seem like a worthwhile investment of time. No strong preference either

src/components/Button/index.js
obedparla: You mean `cx`? Yes indeed

src/components/Button/index.js
obedparla: I was thinking about this yesterday, make it both a standalone exported object and a SC theme, so that we can reuse it easily. I'm getting fed up of not remembering the name of the CSS vars 😅.

I can do that in another PR

src/components/Spinner/index.stories.js
obedparla: agreed, changed it

src/components/Icon/index.js
obedparla: this was there before, kudos to Ugo

.travis.yml
obedparla: I'm gonna remove this. AFAIK we don't use Travis for anything

src/components/UiText/index.js
obedparla: This is close to how Ugo did it right? Yeah I think it's cleaner for components such as UiText, that use a lot of `if`s.

obedparla: For other components I think using `if`s is better. Looking at the `button` there is some logic such as `if (variant === VARIANTS.OUTLINE && intent === INTENTS.PRIMARY)` that would be annoying to implement in object syntax

obedparla: 👍 I really like this when it makes sense: https://github.com/heetch/flamingo/pull/110/commits/5784e8e262def6dce964a97fbf902bb8aba8a29a#diff-4589e5744b98bb9a842c6b69878e29a9

obedparla:
> Nice one! I just had some comments/questions and I'll have a look at the Storybook part when the CI works again

@tamarasaurus The CI won't work in this PR. I have to fix it in another one

src/components/Button/index.js
keanan: Really like the idea of having it as a standalone exported object

PR #111, opened 2020-03-09
Create Dropdown component

ugogo:
This component is inspired by the Popover one
Everything is rendered via render props because we might need to use the `show/hide/toggle` functions
It's recommended to use `Item` as an option, but in the end, you can use whatever you want!

### DIFF
--- a/packages/react/src/components/Dropdown/index.stories.js
+++ b/packages/react/src/components/Dropdown/index.stories.js
@@ -1,0 +1,57 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { action } from '@storybook/addon-actions';
+
+import Dropdown from '.';
+import Button from '../Button';
+import Heading from '../Heading';
+import Item from '../Item';
+
+const stories = storiesOf('Dropdown', module);
+
+stories.add('Playground', () => (
+  <>
+    <Heading>Dropdown</Heading>
+
+    <div
+      style={{
+        display: 'flex',
+        justifyContent: 'center',
+        alignItems: 'baseline',
+        height: '11rem',
+        padding: '1rem',
+        border: '1px dashed var(--f-color-brandSecondary--light)',
+        borderRadius: 'var(--f-borderRadius--m)',
+      }}
+    >
+      <Dropdown
+        options={({ hide }) => (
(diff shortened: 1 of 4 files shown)


ugogo:
THANK YOU OCTAVE

packages/react/src/components/Dropdown/index.stories.js
```diff
@@ -0,0 +1,57 @@
...
+        justifyContent: 'center',
+        alignItems: 'baseline',
+        height: '11rem',
+        padding: '1rem',
+        border: '1px dashed var(--f-color-brandSecondary--light)',
+        borderRadius: 'var(--f-borderRadius--m)',
+      }}
+    >
+      <Dropdown
+        options={({ hide }) => (
```
obedparla: I find this API weird. The children are sent as `options`, and the element that opens the dropdown is in the "children" of the dropdown.

I'd expect the children to be whatever is inside the dropdown instead.

The `Dropdown` component could even be a hook that returns `{ ref, toggle }` and the component itself. The way `react-table` is implemented comes to mind.

I'm kind of thinking out loud here, I'm sure we can have a nicer API for the component, which will be used extensively soon by the care team

ugogo: > I find this API weird. The children are sent as options, and the element that opens the dropdown is in the "children" of the dropdown.
> I'd expect the children to be whatever is inside the dropdown instead.

Ah, yeah, you're right.
But we would still need to use render props instead of a custom hook because we need to wrap some stuff into components, where `react-table` is only focusing on logic

So what about 
```
<Dropdown
  triggerer={({ ref, toggle }) => (
    <Button ref={ref} onClick={toggle}>
      Click me
    </Button>
  )}
>
  {({ hide }) => (
    <>
      <Item onClick={hide}>Batman</Item>
      <Item onClick={hide}>Robin</Item>
    </>
  )}
</Dropdown>
```

obedparla: That's much nicer 👍

ugogo: Fixed @obedparla 👌

PR #117, opened 2020-03-12
v2/move-components

ugogo:
Move actual components into a components-deprecated folder


ugogo:
https://github.com/heetch/flamingo/pull/110

obedparla:
You should also move a copy of `css` into the folder. That would help when migrating since we need to check the stylings etc

ugogo:
@obedparla 

> You should also move a copy of css into the folder. That would help when migrating since we need to check the stylings etc

But we deleted those files 😅

ugogo:
@obedparla Files are still accessible here https://github.com/heetch/flamingo/pull/116/files

obedparla:
🤦‍♂ that's why I deprecated them the way I did. Without being able to access those files I think it's complicated to code. I had the `css` files open on one window and the new SC component on another. 

I wouldn't worry too much about PRs being dirty for now, we're making a lot of changes. WDYT if we add them back? (Sorry I should've mentioned this in the CSS removal PR instead 🤦‍♂ )

PR #120, opened 2020-03-17
Image uploader/fix file detection

ugogo:
Fix a couple of issues leading to weird states and visual glitches


keanan:
Hey @ugo I tested story book and it seems the previous bug is back - if you select the same image twice it no longer shows in the preview.

ugogo:
Hey @keanan, this bug has never been fixed 🤷‍♂ 
But... I just did 🙂

keanan:
It was definitely fixed 🙂 I just double checked again in production

ugogo:
My bad then 🙈

PR #121, opened 2020-03-20
V2/deprecated folders

obedparla:
This PR moves the `css` and `react/components`  into a `deprecated`  folder that will only the used while developing the `V2` version of Flamingo, and removed once we've reached 100% parity.

- Removed usage/mentions of the folders in other files too. 
- Removed `package.json` scripts, they're added in the next PR containing the new base components


package.json
```diff
@@ -10,13 +10,5 @@
   "precommit": "lerna run --concurrency 1 --stream precommit",
   "scripts": {
     "deploy": "lerna bootstrap --npm-client=yarn && lerna publish",
-    "prettier": "prettier --write 'packages/**/*.{js,jsx,json,css,md}'",
-    "css": "yarn --cwd packages/css start",
-    "css:setup": "yarn --cwd packages/css install",
-    "react": "yarn --cwd packages/react start",
```
ugogo: Don't we still need those scripts?

obedparla: They're part of the second PR with the rest of the changes to run the code

obedparla:
CI failing because of so many changes to the codebase

deprecated/components/FileUploader/index.stories.js
```diff
@@ -4,7 +4,7 @@ import { boolean } from '@storybook/addon-knobs';
 import { action } from '@storybook/addon-actions';
 
 import Heading from '../Heading';
-import FileUploader from '.';
```
tamarasaurus: Why change all these if they are now deprecated?

obedparla: This was done automatically by IntelliJ when I moved the files

PR #123, opened 2020-03-23
Add a `header` prop to SidePanel, to use Components in it

obedparla:
- Add a `header` prop to SidePanel, to use Components in it

For a new feature in the admin, we need to be able to send React components to the `side panel` header the way we do it for the footer.

### DIFF
--- a/packages/react/src/components/SidePanel/index.js
+++ b/packages/react/src/components/SidePanel/index.js
@@ -14,6 +14,7 @@ const SidePanel = ({
   isOpen: isOpenProp,
   onClose,
   title,
+  header,
   ...props
 }) => {
   const [isOpen, setIsOpen] = React.useState(isOpenProp);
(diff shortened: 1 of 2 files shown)


packages/react/src/components/SidePanel/index.js
```diff
@@ -51,9 +52,13 @@ const SidePanel = ({
...
         <div className='f-SidePanel-header'>
-          <Heading level={4} as='h1'>
-            {title}
-          </Heading>
+          {header ? (
+            <div className='f-SidePanel-headerComponents'>{header}</div>
+          ) : (
+            <Heading level={4} as='h1'>
+              {title}
+            </Heading>
```
ugogo: What about
```
{ header || <Heading level={4} as='h1'>{title}</Heading>}
```
?

obedparla: Good point! I like it more this way.

PR #126, opened 2020-04-02
In Figma the designers have a "concrete" flamingo color.

obedparla:
Not sure if the name makes sense but we use it quite a lot in the admin and would be nice to have it as a var.


obedparla:
> What is its intent? Is it the default color? In which context is it used?

Elodie uses it mostly for borders, separators, that sort of thing.

ugogo:
So should it be the default color?

obedparla:
Sorry, I don't follow, the default color of what? 😅

ugogo:
Of an icon 😄

PR #130, opened 2020-04-15
Fixs and improvements (Dropdown, Checkbox)

ugogo:


obedparla:
Could you describe what the fix is fixing 🙏 it's hard to review without understanding the context

ugogo:
@obedparla the list is the actual commit list message 😄 
```
Fix Dropdown Overlay z-index
Checkbox should also accept React components as helper and label
Fix document click events
Support rendering dropdown items into a React Portal
```

Let me know if you need more details 🙏

PR #131, opened 2020-04-16
Allow initial state to be passed to Table

tamarasaurus:
This PR updates the `<Table>` component to support passing the initialState prop to define the default sorting on columns.

### DIFF
--- a/packages/react/src/components/Table/index.js
+++ b/packages/react/src/components/Table/index.js
@@ -10,7 +10,15 @@ import RowCell from '../TableRowCell';
 
 const Table = React.forwardRef(
   (
-    { className, columns, data, isSortable, manualSorting, onChangeSort },
+    {
+      className,
+      columns,
+      data,
+      isSortable,
+      manualSorting,
+      onChangeSort,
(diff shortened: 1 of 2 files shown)


packages/react/src/components/Table/index.js
```diff
@@ -99,6 +108,7 @@ Table.propTypes = {
   isSortable: PropTypes.bool,
   manualSorting: PropTypes.bool,
   onChangeSort: PropTypes.func,
+  initialState: PropTypes.shape({}),
```
ugogo, line 111: Is the `initialState` shape always the same? 
If so, could we add a bit more depth?

Like:
```
initialState: PropTypes.shape({
  sortBy: PropTypes.arrayOf(PropTypes.shape({
    id: PropTypes.string.isRequired,
    desc: PropTypes.bool
   }] }
  }))
}),
```

tamarasaurus: It can contain a bunch of other stuff, but I'll update it to support only the `sortBy` for now :+1:

PR #132, opened 2020-04-16
Obed/components migration

obedparla:
A lot of changes here are about changing how previous components were created, by extending components within `styled` and they're now using `attrs` instead. Simpler and more resilient code this way.

- Transition SidePanel component
- Transition Text component
- Transition IconButton

I'm opening this PR to have the code merged before everyone else starts working on other components :)


src/components/Button/index.js
```diff
@@ -97,29 +96,10 @@ const styles = {
...
-      className={cx('f-Button', className)}
-      ref={forwardedRef}
-      children={isLoading ? <Spinner /> : children}
-      {...rest}
-    />
-  ),
-)`
+const Button = styled('button').attrs(({ className, isLoading, children }) => ({
+  className: cx('f-Button', className),
+  children: isLoading ? <Spinner /> : children,
```
ugogo, line 101: I don't think this would work
When a button is loading, we want it to keep the same size, while displaying the loader
That's why we play with elements opacity here
https://github.com/heetch/flamingo/blob/master/packages/react/src/components/Button/index.js#L49-L53

src/components/Button/index.stories.js
```diff
@@ -24,14 +24,14 @@ stories.add('All states', () => (
...
-          <Button intent={intent} style={{ marginRight: 'var(--f-space--m)' }}>
+          <Button intent={intent} margin={'0 var(--f-space--m) 0 0'}>
             {intent}
           </Button>
           <Button
             intent={intent}
             className={`is-${intent}`}
             isLoading
-            style={{ marginRight: 'var(--f-space--m)' }}
+            margin={'0 var(--f-space--m) 0 0'}
```
ugogo, line 34: Is this a good practice?
I wondered if for this kind of stuff using `css` wouldn't be better, wdyt?

src/components/FormComponents/FormElement.js
```diff
@@ -1,7 +1,7 @@
-import React from 'react';
 import styled from 'styled-components';
 import cx from 'classnames';
 import PropTypes from 'prop-types';
+
```
ugogo, line 4: This makes me so happy 😂

src/components/FormComponents/FormElement.js
```diff
@@ -42,9 +42,9 @@ const styles = {
   },
 };
 
-const FormElement = styled(({ className, ...rest }) => (
-  <div className={cx('f-FormEl', className)} {...rest} />
-))`
+const FormElement = styled('div').attrs(({ className }) => ({
+  className: cx('f-FormEl', className),
```
ugogo, line 46: Are the other props automatically spread?

src/components/Heading/index.js
```diff
@@ -17,18 +18,17 @@ const styles = {
     return 'var(--f-space--m)';
   },
 };
-const StyledHeading = styled(({ level, as, className, ...props }) => (
-  <div className={cx('f-Heading', className)} {...props} />
-))`
+
+const Heading = styled(UiText).attrs(({ as, level, className }) => ({
+  as: as || `h${level}`,
```
ugogo, line 23: 👌

src/components/Label/index.js
```diff
@@ -1,19 +1,14 @@
...
-const Label = styled(({ className, htmlFor, ...rest }) => (
-  <UiText
-    variant={UiText.VARIANTS.subContentBold}
-    as='label'
-    className={cx('f-Label', className)}
-    htmlFor={htmlFor}
-    {...rest}
-  />
-))`
+const Label = styled('div').attrs(({ className }) => ({
```
ugogo, line 7:
```suggestion
const Label = styled('label').attrs(({ className }) => ({
```

src/components/SidePanel/index.js
```diff
@@ -0,0 +1,132 @@
...
+  width: 100%;
+  max-width: 33.75rem; /* 540px */
+  overflow: auto;
+  z-index: calc(var(--f-zIndex--Overlay) + 1);
+  transition: transform 0.1s ease-out;
+  background-color: var(--f-color-element--primary);
+  pointer-events: ${({ isOpen }) => (isOpen ? 'inherit' : 'none')};
+  transform: ${({ isOpen }) => (isOpen ? 'translateX(0)' : 'translateX(100%)')};
+
+  @media (max-width: 460px) {
```
ugogo, line 100: Should we create constants for that?
Like `onMobile`, and the we'll be able to do stuff like
```
`${onMobile}` {
   ...
}
```

src/components/SidePanel/index.js
```diff
@@ -0,0 +1,132 @@
...
+
+const Footer = styled('div')`
+    position: sticky;
+    bottom: 0;
+    display: flex;
+    justify-content: flex-end;
+    border-top: 1px solid var(--f-color-element--inactive);
+    padding: 0 var(--content-hSpacing);
+
+   ${Button} + ${Button} {
```
ugogo, line 128: Shouldn't this go into the `Button` styles?

src/components/Text/index.js
```diff
@@ -0,0 +1,31 @@
...
+const variants = Object.values(UiText.VARIANTS);
+
+const Text = styled(UiText).attrs(({ className }) => ({
+  className: cx('f-Text', className),
+}))`
+  color: var(--f-color-text--secondary);
+  text-align: ${({ isNumber }) => isNumber && 'right'};
+`;
+
+Text.displayName = 'Text';
```
ugogo, line 16: Should we add a babel plugin to automatically do that?

src/components/Button/index.stories.js
obedparla: I was wondering the same thing, the reason I went for this is that `css` will "extend" and add extra css, so there will be 2 `margin`, plus you need to use the styled-components babel plugin for `css` to work.

It's not a big deal, either way, just thought this would be slightly easier 🤷‍♂️

src/components/FormComponents/FormElement.js
obedparla: 😂 copy pasting them clases

src/components/FormComponents/FormElement.js
obedparla: Yup. `attrs` kinda extends it.

src/components/Label/index.js
obedparla: 👌 good catch

src/components/SidePanel/index.js
obedparla: That's a great idea. I'm going to do this on another PR 👍

src/components/Text/index.js
obedparla: 👌 will do that

src/components/SidePanel/index.js
obedparla: I added it this way because that's what we had in Flamingo. We could add it to `Button` but that would be a styling difference from our current implementation

src/components/Button/index.js
obedparla: You're right, this changes the behavior completely. Will fix and add an extra story to it

src/components/Button/index.stories.js
ugogo: 👌

package.json
```diff
@@ -35,7 +35,6 @@
   },
   "devDependencies": {
     "@babel/core": "^7.7.2",
-    "@babel/plugin-proposal-class-properties": "^7.7.0",
```
ugogo, line 38: You removed it here but it's still present in the babelrc file, is it expected?
https://github.com/heetch/flamingo/pull/132/commits/c78d737bad16f82a85815504b059ffb08725c508#diff-e56633f72ecc521128b3db6586074d2cR4

obedparla: good catch 😅

PR #134, opened 2020-04-20
searchPlusIcon

keanan:
Add search icon with plus sign

image


packages/react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@heetch/flamingo-react",
-  "version": "1.15.2",
```
obedparla, line 3: Shouldn't this be done after merging, and by Lerna automatically?

ugogo: Yes @obedparla, we must not update manually any version 🙏

lerna.json
```diff
@@ -3,5 +3,5 @@
   "packages": [
     "packages/*"
   ],
-  "version": "1.15.2"
+  "version": "1.15.3"
```
ugogo, line 6: Same here https://github.com/heetch/flamingo/pull/134/files#r411548056

packages/react/package.json
keanan: Yes you are right, thanks

PR #135, opened 2020-04-20
Obed/components migration 2

obedparla:
Opening another PR with the following components, I'm doing it in batches to prevent a massive PR with all the components migrated. This includes:

- - SidePanel rendering rehaul to animate on mount and unmount. I hope this doesn't break any current functionality, but it will enable much better usage of the component.
- Text 
- Link
- Textarea & TextareaField
- IconButton
- UploaderItem
- UploaderImageItem
- ImageUploader & ImageUploaderField
- FileUploader & FileUploaderField

Remarks:

- I removed some classes that were "is-loading", "is-error" and so on, since they were used for styling purposes and I don't think we'll need to use them in external products
- Removed default props for `undefined` values. I think it's pointless to give a default of undefined. The rule is disabled too


src/components/FileUploader/styles.js
```diff
@@ -0,0 +1,86 @@
...
+  background-color: var(--f-color-element--primary);
+  transition: all 0.15s ease-out;
+  border: 1px dashed ${styles.borderColor};
+  height: ${({ hasFile }) =>
+    hasFile ? 'auto' : 'var(--f-FileUploader-height)'};
+  text-align: ${({ hasFile }) => hasFile && 'center'};
+
+  ${({ hasFile }) =>
+    !hasFile &&
+    css`
```
ugogo, line 44: I don't think you have to wrap this into `css`?

src/components/FileUploader/index.js
```diff
@@ -0,0 +1,195 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+
+import Icon from '../Icon';
+import Spinner from '../Spinner';
+import UploaderItem from '../UploaderItem';
+import { StyledFileUploader, FileUploaderLabel } from './styles';
```
ugogo, line 8: Why is the first component imported prefixed by `Styled` when the second is not?

src/components/FileUploader/styles.js
```diff
@@ -0,0 +1,86 @@
...
+    if (invalid) {
+      return 'var(--f-color-element--error)';
+    }
+    if (valid) {
+      return 'var(--f-color-element--success)';
+    }
+    return 'var(--f-color-text--secondary)';
+  },
+  borderColor({ invalid, valid, hasFile }) {
+    if (hasFile) {
```
ugogo, line 16: Could we use onliners here?
IMO it would make to whole file more readable

```
borderColor({ invalid, valid, hasFile }) {
    if (hasFile) return 'transparent'
    if (invalid) return 'var(--f-color-element--error)';
    if (valid) return 'var(--f-color-element--success)';
    return 'var(--f-color-element--tertiary)';
  },
```

src/components/FileUploader/styles.js
```diff
@@ -0,0 +1,86 @@
...
+
+  &:hover {
+    color: var(--f-color-brandPrimary);
+  }
+
+  &.f-FileUploader-state--addFiles {
+    padding: var(--f-space--l);
+  }
+
+  .f-Icon {
```
ugogo, line 82:
```suggestion
  ${Icon} {
```

src/components/FileUploader/styles.js
```diff
@@ -0,0 +1,86 @@
...
+    margin-bottom: ${({ withIconMargin }) =>
```
ugogo, line 83: What about?
```suggestion
    margin-bottom: ${({ withIcon }) =>
```

src/components/FileUploaderField/index.js
```diff
@@ -0,0 +1,26 @@
...
+FileUploaderField.propTypes = {
+  helper: PropTypes.node,
+  id: PropTypes.string.isRequired,
+  label: PropTypes.node,
+  inputRef: PropTypes.shape({
+    current: PropTypes.instanceOf(Element),
+  }),
+};
+
+FileUploaderField.displayName = 'FileUploaderField';
```
ugogo, line 24: `displayName` is not needed anymore IIRC 🙂

src/components/ImageUploader/index.js
```diff
@@ -0,0 +1,126 @@
...
+        {...props}
+      >
+        {hasPreview && (
+          <>
+            <UploaderPreview
+              className='f-ImageUploader-preview'
+              preview={preview}
+            />
+
+            <div className='f-ImageUploader-hoverState'>
```
ugogo, line 84: Shouldn't this be a component by itself?

src/components/SidePanel/index.js
```diff
@@ -69,64 +78,17 @@ SidePanel.propTypes = {
   closesOnOverlayClick: PropTypes.bool,
   footer: PropTypes.node,
   isOpen: PropTypes.bool,
-  animateOnMount: PropTypes.bool,
   onClose: PropTypes.func,
   title: PropTypes.node,
   header: PropTypes.node,
+  maxWidth: PropTypes.string,
```
ugogo, line 84: Leftover?

src/components/SidePanel/index.js
```diff
@@ -1,45 +1,54 @@
...
 const SidePanel = props => {
   const { footer, header } = props;
 
-  // animateOnMount makes the panel closed by default so that on the first render
-  // It is set to props.isOpen (through the useEffect)
-  // if props.isOpen is true, then it will animate <3
-  const [isOpen, setIsOpen] = React.useState(
-    props.animateOnMount ? false : props.isOpen,
-  );
+  const [shouldRender, setShouldRender] = React.useState(props.isOpen);
```
ugogo, line 18: What is the difference between this and having a condition in the render?
Like `isOpen && ...`

src/components/UiText/index.js
```diff
@@ -70,15 +70,18 @@ const UiText = styled('p').attrs(({ variant }) => ({
   font-weight: ${styles.fontWeight};
   font-size: ${styles.fontSize};
   line-height: ${styles.lineHeight};
+  margin: ${({ margin }) => margin};
+  color: ${({ color }) => color};
 `;
 
 UiText.propTypes = {
   className: PropTypes.string,
+  margin: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
```
ugogo, line 79: I find this so weird 😞

ugogo: Since props are spread, maybe there is no need for declaring every "styling propTypes"?

src/components/FileUploader/styles.js
obedparla: True, but without it we don't get syntax highlight 😅. I don't think it makes any difference once compiled

src/components/FileUploader/index.js
obedparla: I can't name it FileUploader since that's the name of the component itself. We used to prefix `Styled` to everything in the `admin` before and that got annoying real fast, so now we do it only when necessary.

I'd be happy to change it to any other name if it makes more sense

src/components/FileUploader/styles.js
obedparla: I didn't think of it, looks a lot better! 👍

src/components/FileUploaderField/index.js
obedparla: dammit I keep forgetting 😄

src/components/FileUploader/styles.js
obedparla: yasss boi ( ͡° ͜ʖ ͡°)

src/components/ImageUploader/index.js
obedparla: Do you know if you're using these classes anywhere? I'd like to remove them but I'm not sure if they're being used. We're not using `f-ImageUploader-hoverState` or `f-ImageUploader-preview` in the admin.

They're not that important, if I can remove them it's cleaner as a SC

src/components/SidePanel/index.js
obedparla: it's passed to `StyledSidePanel`. Changed it to pass it explicitly.

src/components/UiText/index.js
obedparla: I don't like it either, but it always bites me in the ass because one tends to pass `margin={0}` a lot.

If we don't declare it as a prototype then an outside product wouldn't know what props it accepts

src/components/SidePanel/index.js
obedparla: The `isOpen` prop controls the animation. `shouldRender` state controls the rendering. isOpen state is used to be able to close on click of the X or the overlay rather than by prop.

If I use `isOpen` in the render directly, it would hide it instantly without any animation, I need a flag to know when the animation is finished _and then_ hide the element.

Not sure if I'm clear. I tried to explain it in a comment in the file

src/components/FileUploaderField/index.js
obedparla: I checked and we need it here since it's not an SC but a normal component

src/components/FileUploader/styles.js
obedparla: https://github.com/heetch/flamingo/pull/135/commits/1168170e55ce7c172e152c368afd21c7c5ae84c4

You inspired me 😄

ugogo: 😍

src/components/FileUploader/index.js
ugogo: No ok, I was just wondering 🙂

src/components/FileUploader/styles.js
ugogo: 💃

src/components/FileUploaderField/index.js
ugogo: Oh yeah of course 🤦

src/components/SidePanel/index.js
ugogo: Okayyy 👍

src/components/UiText/index.js
ugogo: Dunno what's best here, because it also means that if we want to add another prop like this, we will have to write another proptype and release a new version, every time 🤔

obedparla: Yeah, but that makes sense to me, we'd do the same if we added a normal prop.

We're not covering all uses since we have `css`, only the ones that are so common they're worth having as a prop.

ugogo: I guess that makes sense yes
@tamarasaurus @keanan wdyt?

obedparla: @OctaveStoufflet since you're the only one that seems to check his Github notifications 😄

OctaveStoufflet: As far as I understood, if we tend to pass `margin={0}` it make sense to pass it as a props, since it's more "practical" ? But... if we "always" wanna have margin = 0, why didn't we write directly
`margin : 0` in the styled rules ?

OctaveStoufflet: well not "always" maybe.. but if we tend to have `margin: 0` we can choose to have `margin: 0` as a styled by default. and, if we wanna change this rules wile we are working on something, we can change the rules in the `styles.jsx` of the Component we are working on, no ?

OctaveStoufflet: side note : thank you @ugogo i'm very honored to see that my insights matter.

obedparla: You're right :) we don't want `margin=0` to be the default, but it's common enough to be a prop on its own.

The question is mostly wether we should list all `style` props in `propTypes`, which I voted for

ugogo: Oh no @OctaveStoufflet sorry 😭 ❤️

OctaveStoufflet: oooooh ok didn't understood the question sorry 
Yes we should indeed, to be aware

PR #136, opened 2020-04-21
Styled components/add theme support

ugogo:


src/components/Button/index.js
```diff
@@ -128,13 +98,13 @@ const Button = styled(
...
-  margin: ${({ margin }) => margin ?? 'var(--f-space--m) 0'};
-  padding: var(--f-space--m) var(--f-space--xl);
-  font-size: var(--f-fontSize--s);
-  font-weight: var(--f-fontWeight--bold);
-  line-height: var(--f-lineHeight--m);
+  margin: ${({ margin, theme }) => margin ?? `${theme.spaces.m} 0`};
+  padding: ${({ theme }) => `${theme.spaces.m} ${theme.spaces.xl}`};
+  font-size: ${({ theme }) => theme.fontSizes.s};
+  font-weight: ${({ theme }) => theme.fontWeights.bold};
+  line-height: ${({ theme }) => theme.lineHeights.m};
```
obedparla, line 105: I feel strongly against this, it would force us to always wap all our apps in a `<FlamingoThemeProvider>`, otherwise the components will explode.

In the admin we have many mini-apps, so that'd be annoying.

We should instead have a theme as a `js` object and import it whenever we need it. That way we don't have to wrap our apps with a theme provider.

We can still export a  `FlamingoTheme`, but we won't be forced to use it

src/constants/index.js
```diff
@@ -5,4 +5,89 @@ export const refShapes = [
...
+    text: {
+      primary: 'var(--f-color-text--primary)',
+      secondary: 'var(--f-color-text--secondary)',
+      tertiary: 'var(--f-color-text--tertiary)',
+      link: 'var(--f-color-text--link)',
+      white: 'var(--f-color-text--white)',
+    },
+  },
+
+  fontSizes: {
```
obedparla, line 40: I think it's more intuitive to use a singular name: `fontSize`, `space` `fontWeight` rather than the plural, same as with the ones above (text, element, icon, etc). Then we use it as `theme.fontSize.xs`, rather than `theme.fontSizes.xs` and so on.

src/constants/index.js
```diff
@@ -5,4 +5,89 @@ export const refShapes = [
   PropTypes.shape({ current: PropTypes.instanceOf(Element) }),
 ];
 
-export default { refShapes };
+export const theme = {
```
obedparla, line 8: WDYT about moving theme to a file on its own?

src/components/Button/index.js
ugogo: I don't think that would be a good idea, what's the point of using a theme (especially to be future-proof) if we don't use it?
Also, wrapping an app into a Provider shouldn't be that complicated given all the benefits you'll have after that (think using the `useTheme` hook for example)
I'm obviously not in your day-to-day work so bear with me, but f we want to add consistency in our apps, having and using a theme seems to me the way to go

src/constants/index.js
ugogo: Good point 👍

src/constants/index.js
ugogo: Done 👍

src/components/Button/index.js
obedparla: What I meant was to still use the `theme` throughout Flamingo, but not directly from styled component props, but by importing it `import theme from '../theme'`.

Using the theme from props is useful when the theme might change (dark mode, light mode sort of thing). But since ours is static we can use it by importing it, and that way we don't have a dependency on the `ThemeProvider` when using Flamingo in a product.

This has all the benefits IMO, simpler code, but we can also export a `FLamingoThemeProvider` that we can use if we want, giving us access to `useTheme`. And if we don't want to use the theme provider Flamingo would still work.

Plus we can import the theme object directly in a product if we need to. 

WDYT?

keanan: +1 To Obed's comment, don't think we should use `ThemeProvider` where we don't need to, rather import when we do. `ThemeProvider` adds wrapper code to each component, so we should avoid using it where not required.

obedparla:
@ugogo I'm taking over this PR since I'm done migrating components

ugogo:
Thanks @obedparla :mvp:

PR #139, opened 2020-04-22
Obed/components migration 3

obedparla:
Migrates:

- Item
- Dropdown
- Card
- Table
    - TableHeaderCell
    - TableRowCell

Removes (see a comment in the code)
    - TableHeaderGroup
    - TableRowGroup

Fix a few other things.


src/components/Button/index.js
```diff
@@ -142,6 +131,10 @@ const Button = styled(
   background-color: ${styles.backgroundColor};
   box-shadow: ${styles.boxShadow};
 
+  & + & {
+    margin-left: var(--f-space--m);
+  }
+
```
obedparla, line 137: @ugogo I added this here. It's a change in how the component behaves but I think it makes sense. WDYT?

src/components/Table/index.js
```diff
@@ -0,0 +1,131 @@
...
+Table.defaultProps = {
+  columns: [],
+  data: [],
+  isSortable: false,
+  onChangeSort: () => {},
+  manualSorting: false,
+};
+
+Table.HeaderCell = HeaderCell;
+Table.RowCell = RowCell;
```
obedparla, line 129: Originally we were exporting 

```
Table.HeaderGroup = HeaderGroup;
Table.HeaderCell = HeaderCell;
Table.RowGroup = RowGroup;
Table.RowCell = RowCell;
```

But both `HeaderGroup` and `RowGroup` are just an HTML element receiving ref, classname, props, children. All of what is done automatically by an element.

I didn't see the reason in exporting them either, since they don't have any functionality. 

Just want to confirm you're not using it anywhere in your apps for some reason @ugogo

keanan:
Hey Obed, thanks for doing this, amazing job! 

Not sure if we doing this in the new SC but if it is possible it would be really nice for us to allow sorting per column and not the entire table, also if we could add tooltips for the header cells that would be cool, they currently show `Toggle SortBy` by default - just a suggestion if you have time thanks.

obedparla:
> Hey Obed, thanks for doing this, amazing job!
> 
> Not sure if we doing this in the new SC but if it is possible it would be really nice for us to allow sorting per column and not the entire table, also if we could add tooltips for the header cells that would be cool, they currently show `Toggle SortBy` by default - just a suggestion if you have time thanks.

Hey buddy, I've been improving some components here and there, I didn't plan to add this functionality. I want to finish and publish this week if possible. 

If I have time I'll check on this. But I have a question: how are we going to sort per column? Each row represents one piece of data. Sorting a column only and ignoring the rest would create a discrepancy of data and wouldn't make any sense, not sure if I'm missing the point 🤔

tamarasaurus:
@obedparla I think he means that right now you can define the whole table is sortable or not, but we want to do it on a per-column basis. That way we could set some columns as sortable without having to set `accessor` to null or doing some tricks. :heavy_plus_sign: that would be great to have

keanan:
Hey man, all good thanks was just a small suggestion, I will hopefully have time soon to add some stuff (like a pagination element) - sorry I was not being clear - if we can disable sorting for certain columns is what I meant. So columns which have un-sortable data (like icons, buttons etc) should not be able to sort.

keanan:
Thanks Tam, beat me to it 🙂

obedparla:
I'll see if I can implement it, it makes sense to have that.

Also: maybe review it now that you're here? 😢

src/components/Card/index.js
```diff
@@ -0,0 +1,66 @@
...
+
+const styles = {
+  padding({ size }) {
+    if (size === 's') return 'var(--f-space--xl)';
+    if (size === 'm') return 'var(--f-space--xxl)';
+
+    return undefined;
+  },
+  boxShadow({ isSelected, elevation }) {
+    if (elevation === 2) return '0 4px 20px rgba(25, 1, 52, 0.14)';
```
tamarasaurus, line 14: do we have those box shadows in the theme?

src/components/Dropdown/index.js
```diff
@@ -0,0 +1,122 @@
...
+
+import Overlay from '../Overlay';
+import { StyledDropdown } from './styles';
+
+const Dropdown = ({
+  children,
+  clickOutsideToHide,
+  onHide,
+  onOpen,
+  portalNode,
```
tamarasaurus, line 14: why not include `modifiers` here?

src/components/Item/styles.js
```diff
@@ -0,0 +1,73 @@
...
+import PropTypes from 'prop-types';
+import Icon from '../Icon';
+
+const styles = {
+  color({ size }) {
+    if (size === 'mini') return 'var(--f-color-text--secondary)';
+
+    return 'var(--f-color-text--primary)';
+  },
+  xPadding({ size }) {
```
tamarasaurus, line 11: why xPadding? just curious about the naming :)

src/components/SidePanel/index.js
```diff
@@ -49,6 +49,7 @@ const SidePanel = props => {
         onAnimationEnd={onAnimationEnd}
         maxWidth={props.maxWidth}
         isOpen={isOpen}
+        animateOnMount={props.animateOnMount}
```
tamarasaurus, line 52: :+1:

src/components/Table/index.js
```diff
@@ -0,0 +1,131 @@
...
+);
+
+Table.displayName = 'Table';
+
+Table.propTypes = {
+  className: PropTypes.string,
+  columns: PropTypes.arrayOf(
+    PropTypes.shape({
+      Header: PropTypes.string,
+      accessor: PropTypes.string.isRequired,
```
tamarasaurus, line 103: Could we have this as not required until we can sort columns individually?

src/components/Card/index.js
```diff
@@ -0,0 +1,66 @@
...
+    border-radius: var(--f-borderRadius--l);
+  }
+
+  & & & {
+    border-radius: var(--f-borderRadius--m);
+  }
+
+  & & & & {
+    border-radius: var(--f-borderRadius--s);
+  }
```
keanan, line 40: are these overrides? how do they work?

src/components/Item/styles.js
keanan: its used for top/bottom so maybe `yPadding` 😄

src/components/Button/index.js
ugogo: Yep LGTM 👍

src/components/Table/index.js
ugogo: I think it's if we want to render a table piece by piece, not sure if that's used though

src/components/Card/index.js
obedparla: nope but I'll add them!

src/components/Dropdown/index.js
obedparla: just to save 1 like of code 😅. I like the explicitness of `props.something` and it saves space xD

src/components/Item/styles.js
obedparla: ops. It should be `yPadding` indeed. Updated.

y axis = horizontal. Is it too explicit?

src/components/Table/index.js
obedparla: Sure 👍

src/components/Card/index.js
obedparla: Each `&` is the class of the component. So this would be

```.myClass.myClass.myClass.myClass```

The original Card had this with classes so I had to reproduce it like this.

PR #140, opened 2020-04-22
Allow disabled select option

ugogo:


obedparla:
Why don't we allow to send any props to it instead?

ugogo:
There is no more prop to spread @obedparla 😄 
https://www.w3schools.com/tags/tag_option.asp

PR #141, opened 2020-04-23
Datepicker component

tamarasaurus:
This PR adds a new datepicker component based on react-day-picker (https://bundlephobia.com/result?p=react-day-picker@7.4.8)

### DIFF
--- a/packages/css/src/components/Datepicker.css
+++ b/packages/css/src/components/Datepicker.css
@@ -1,0 +1,53 @@
+.f-Datepicker {
+  position: relative;
+  cursor: pointer;
+  text-align: left;
+}
+
+.f-Datepicker .DayPicker-Weekday {
+  color: var(--f-color-element--tertiary);
+}
+
+.f-Datepicker .DayPicker {
+  display: flex;
+  justify-content: center;
+}
+
+.f-Datepicker--wrapper {
+  width: 100%;
+}
+
+.f-Datepicker .DayPicker-Day--today {
+  color: var(--f-color-brandPrimary);
+}
+
+.f-Datepicker
+  .DayPicker:not(.DayPicker--interactionDisabled)
+  .DayPicker-Day:not(.DayPicker-Day--disabled):not(.DayPicker-Day--selected):not(.DayPicker-Day--outside):hover {
+  background-color: var(--f-color-element--tertiary);
+}
+
+.f-Datepicker
+  .DayPicker-Day--selected:not(.DayPicker-Day--disabled):not(.DayPicker-Day--outside) {
+  background-color: var(--f-color-brandPrimary);
+}
+
+.f-Datepicker--navbar {
+  padding-left: var(--f-space--xxl);
+  padding-right: var(--f-space--xl);
+  padding-top: var(--f-space--xl);
+  color: var(--f-color-text--primary);
+  font-weight: var(--f-fontWeight--bold);
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+}
+
+.f-Datepicker--actions .f-Button {
+  margin: 0;
+  cursor: pointer;
+}
+
+.f-Datepicker .DayPicker-Month {
+  margin-top: 0;
+}
--- a/packages/react/src/components/Datepicker/index.js
+++ b/packages/react/src/components/Datepicker/index.js
@@ -1,0 +1,88 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import DayPicker from 'react-day-picker';
+
+import 'react-day-picker/lib/style.css';
+import DatepickerNavigation from '../DatepickerNavigation';
+import DatepickerInput from '../DatepickerInput';
+import Dropdown from '../Dropdown';
+
+const Datepicker = ({
+  className,
+  invalid,
+  valid,
+  placeholder,
+  value,
+  onChange,
+  formatDate,
+  locale,
+  renderNavigation,
+}) => {
+  return (
+    <div
+      className={cx('f-FormEl-wrapper f-Datepicker', {
+        'is-invalid': invalid,
+        'is-valid': valid,
+        className,
+      })}
+    >
+      <Dropdown
+        placement='bottom-start'
+        triggerer={({ ref, toggle }) => (
+          <DatepickerInput
+            ref={ref}
+            formatDate={formatDate}
+            value={value}
+            onClick={toggle}
+            invalid={invalid}
+            placeholder={placeholder}
+          />
+        )}
+      >
+        {({ hide }) => (
+          <DayPicker
+            value={value}
+            locale={locale}
+            selectedDays={[value]}
+            initialMonth={value}
+            onDayClick={date => {
+              onChange(date);
+              hide();
+            }}
+            captionElement={() => null}
+            navbarElement={renderNavigation}
+          />
+        )}
+      </Dropdown>
+    </div>
+  );
+};
+
+Datepicker.displayName = 'Datepicker';
+
+Datepicker.propTypes = {
+  className: PropTypes.string,
+  invalid: PropTypes.bool,
+  valid: PropTypes.bool,
+  placeholder: PropTypes.string,
+  value: PropTypes.instanceOf(Date),
+  onChange: PropTypes.func,
+  formatDate: PropTypes.func,
+  renderNavigation: PropTypes.func,
+  locale: PropTypes.string,
+};
+
+Datepicker.defaultProps = {
+  placeholder: 'Select a date',
+  invalid: false,
+  valid: false,
+  className: undefined,
+  value: null,
+  locale: 'en',
+  formatDate: date => date.toLocaleDateString('en-GB'),
+  onChange: () => {},
+  renderNavigation: props => <DatepickerNavigation {...props} />,
+};
+
+export default Datepicker;
--- a/packages/react/src/components/DatepickerInput/index.js
+++ b/packages/react/src/components/DatepickerInput/index.js
@@ -1,0 +1,41 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Icon from '../Icon';
+
+const DatepickerInput = React.forwardRef(
+  ({ onClick, placeholder, value, invalid, formatDate }, ref) => (
+    <button
+      ref={ref}
+      type='button'
+      onClick={onClick}
+      onKeyPress={({ which }) => (which === 13 ? onClick() : undefined)}
+      className='f-FormEl f-FormEl--withIcon f-Datepicker'
+    >
+      {value ? formatDate(value) : placeholder}
+      <Icon
+        icon={invalid ? Icon.ICONS.IconAlertOctagon : Icon.ICONS.IconCalendar}
+        className='f-FormEl-icon'
+      />
+    </button>
+  ),
+);
+
+DatepickerInput.displayName = 'DatepickerInput';
+
+DatepickerInput.propTypes = {
+  onClick: PropTypes.func,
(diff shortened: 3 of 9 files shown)


packages/react/src/components/Datepicker/index.js
```diff
@@ -0,0 +1,95 @@
...
+        className='f-FormEl-icon'
+      />
+    </div>
+  ),
+);
+
+const Datepicker = ({ className, isInvalid, placeholder, value, onChange }) => {
+  return (
+    <div
+      className={cx('f-FormEl-wrapper f-Datepicker', {
```
ugogo, line 49: Why not using a Dropdown here?
With it, we would benefit from the "clicking outside will hide it" feature and improve the mobile display

packages/react/src/components/Datepicker/index.js
```diff
@@ -0,0 +1,95 @@
...
+        'is-invalid': isInvalid,
+      })}
+    >
+      <DayPickerInput
+        onDayChange={date => onChange(date)}
+        formatDate={date => date.toLocaleDateString()}
```
ugogo, line 55: This should be moved into props to improve the customization of the component

packages/react/src/components/Datepicker/index.js
```diff
@@ -0,0 +1,95 @@
...
+        placeholder={placeholder}
+        classNames={{
+          container: 'DayPickerInput f-Datepicker--wrapper',
+          overlay: 'f-Datepicker--overlay',
+        }}
+        formatDate={date => date.toLocaleDateString(['en-GB'])}
```
ugogo, line 61: This should be moved into props to improve the customization of the component

packages/react/src/components/Datepicker/index.js
```diff
@@ -0,0 +1,95 @@
...
+        inputProps={{ isInvalid }}
```
ugogo, line 62:
```suggestion
        inputProps={{ isInvalid, isValid }}
```

packages/react/src/components/Datepicker/index.js
```diff
@@ -0,0 +1,95 @@
...
+          />
+        </div>
+      </div>
+    );
+  },
+);
+
+const DatepickerInput = React.forwardRef(
+  ({ onClick, placeholder, value, isInvalid }, ref) => (
+    <div
```
ugogo, line 32: Interesting, why not using an Input?

packages/react/src/components/Datepicker/index.js
```diff
@@ -0,0 +1,95 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import DayPickerInput from 'react-day-picker/DayPickerInput';
+import 'react-day-picker/lib/style.css';
+import Icon from '../Icon';
+import IconButton from '../IconButton';
+
+const DatepickerNavigation = React.forwardRef(
```
ugogo, line 9: Could this be moved into a prop or something?
For example, in the driver signup, we will want to render another component because we want to be able to select the month and the year

packages/react/src/components/Datepicker/index.js
tamarasaurus: For the design that I was implementing this component is just for display, it's not editable

packages/react/src/components/Datepicker/index.js
tamarasaurus: I just used the in-built hide/show of react-day-picker. I can use the dropdown instead :+1:

packages/react/src/components/Datepicker/index.js
tamarasaurus: @ugogo Done, let me know if it's ok :slightly_smiling_face:

packages/react/src/components/Datepicker/index.js
tamarasaurus: Done

packages/react/src/components/Datepicker/index.js
```diff
@@ -0,0 +1,72 @@
...
+  onChange: PropTypes.func,
+  formatDate: PropTypes.func,
+  NavbarElement: PropTypes.oneOfType([PropTypes.func, PropTypes.object]),
+};
+
+Datepicker.defaultProps = {
+  placeholder: 'Select a date',
+  isInvalid: false,
+  className: undefined,
+  value: null,
```
obedparla, line 68: can you remove these? I'm removing all non-necessary defaultProps for V2.

packages/react/src/components/DatepickerInput/index.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Icon from '../Icon';
+
+const DatepickerInput = React.forwardRef(
+  ({ onClick, placeholder, value, isInvalid }, ref) => (
+    <div
+      role='button'
+      tabIndex={0}
```
obedparla, line 9: It should be a `button` instead, better semantics than a div with role button. You may need to add two more lines of CSS to make it match but I'm doing those changes where I can for V2

packages/react/src/components/DatepickerInput/index.js
```diff
@@ -0,0 +1,40 @@
...
+  placeholder: PropTypes.string,
+  value: PropTypes.string,
+  isInvalid: PropTypes.bool,
+};
+
+DatepickerInput.defaultProps = {
+  onClick: PropTypes.func,
+  placeholder: 'Select a date',
+  value: null,
+  isInvalid: false,
```
obedparla, line 37: Please remove these too 🙏

packages/react/src/components/DatepickerInput/index.js
```diff
@@ -0,0 +1,40 @@
...
+
+DatepickerInput.propTypes = {
+  onClick: PropTypes.func,
+  placeholder: PropTypes.string,
+  value: PropTypes.string,
+  isInvalid: PropTypes.bool,
+};
+
+DatepickerInput.defaultProps = {
+  onClick: PropTypes.func,
```
obedparla, line 34: I don't think PropTypes work inside defaultProps 😄. You can remove it since it' not necessary here

packages/react/src/components/DatepickerNavigation/index.js
```diff
@@ -0,0 +1,41 @@
...
+        {month.getFullYear()}
+        <div className='f-Datepicker--actions'>
+          <IconButton
+            onClick={() => onPreviousClick()}
+            icon={Icon.ICONS.IconChevronLeft}
+          />
+          <IconButton
+            onClick={() => onNextClick()}
+            icon={Icon.ICONS.IconChevronRight}
+          />
```
obedparla, line 20:
```suggestion
          <IconButton
            onClick={onPreviousClick}
            icon={Icon.ICONS.IconChevronLeft}
          />
          <IconButton
            onClick={onNextClick}
            icon={Icon.ICONS.IconChevronRight}
          />
```

packages/react/src/components/DatepickerNavigation/index.js
```diff
@@ -0,0 +1,41 @@
...
+  month: PropTypes.instanceOf(Date),
+  onPreviousClick: PropTypes.func,
+  onNextClick: PropTypes.func,
+};
+
+DatepickerNavigation.defaultProps = {
+  month: null,
+  onPreviousClick: () => {},
+  onNextClick: () => {},
+};
```
obedparla, line 39: Can be removed.

Also, if `month` is null the component would explode since you're trying to run methods on it. Maybe `if(!month) return null`

packages/css/src/components/Datepicker.css
```diff
@@ -0,0 +1,50 @@
...
+  box-shadow: 0 2px 10px rgba(25, 1, 52, 0.08);
+}
+
+.f-Datepicker .DayPicker-Day--today {
+  color: var(--f-color-brandPrimary);
+}
+
+.f-Datepicker
+  .DayPicker:not(.DayPicker--interactionDisabled)
+  .DayPicker-Day:not(.DayPicker-Day--disabled):not(.DayPicker-Day--selected):not(.DayPicker-Day--outside):hover {
```
obedparla, line 23: This is gonna be so much fun to migrate to styled components 😅

obedparla, line 23: This is gonna be so much fun to migrate to styled components 😅

packages/react/src/components/DatepickerNavigation/index.js
tamarasaurus: That won't work because of the way the library works

packages/react/src/components/Datepicker/index.js
```diff
@@ -0,0 +1,72 @@
...
+          container: 'DayPickerInput f-Datepicker--wrapper',
+          overlay: 'f-Datepicker--overlay',
+        }}
+        formatDate={formatDate}
+        inputProps={{ isInvalid }}
+        component={DatepickerInput}
+        value={value}
+        dayPickerProps={{
+          value,
+          locale: 'fr',
```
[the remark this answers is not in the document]
tamarasaurus: Already fixed, it's now a prop :+1:

packages/react/src/components/Datepicker/index.js
```diff
@@ -0,0 +1,72 @@
...
+}) => {
+  return (
+    <div
+      className={cx('f-FormEl-wrapper f-Datepicker', {
+        'is-invalid': isInvalid,
+        className,
+      })}
+    >
+      <DayPickerInput
+        locale='fr'
```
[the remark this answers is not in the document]
tamarasaurus: Already fixed, it's now a prop :+1:

packages/react/src/components/Datepicker/index.js
```diff
@@ -0,0 +1,72 @@
...
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import DayPickerInput from 'react-day-picker/DayPickerInput';
+import 'react-day-picker/lib/style.css';
+import DatepickerNavigation from '../DatepickerNavigation';
+import DatepickerInput from '../DatepickerInput';
+
+const Datepicker = ({
+  className,
+  isInvalid,
```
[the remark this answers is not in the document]
tamarasaurus: Sure, good point

packages/react/src/components/DatepickerInput/index.js
tamarasaurus: We still have `react/require-default-props` as an eslint rule so I'll keep the default props for now

packages/react/src/components/DatepickerInput/index.js
tamarasaurus: Fixed :+1:

packages/react/src/components/Datepicker/index.js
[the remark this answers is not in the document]
tamarasaurus: Fixed

packages/react/src/components/Datepicker/index.js
tamarasaurus: Fixed

packages/react/src/components/DatepickerInput/index.js
obedparla: Ah right. I removed that rule for V2. I'll remove the defaults in the migration then

tamarasaurus:
@ugogo I update the component to use `Dropdown`. For the i18n I'm using the `locale` prop passed into the datepicker and using it to display the month localised. For the display of the selected date, it uses `formatDate` which can also be passed in as a prop too. 

@obedparla I fixed your comments minus the default prop types - `react/require-default-props` is still activated so I'll leave it for now.

tamarasaurus:
Design QA passed :heavy_check_mark:

PR #144, opened 2020-04-24
Add theme support

obedparla:
- Create a theme (copied from https://github.com/heetch/flamingo/pull/136)
- Replace all Flamingo vars with the theme. 
- Replace media queries for the theme media queries 🎉 
- Export the theme so external products can use it instead of the flamingo vars

### The file changes are massive, but it's basically a find-replace-all of the vars for the theme counterpart. I double-checked every single file and tested all components again to make sure everything still worked 😄


src/theme.js
```diff
@@ -0,0 +1,79 @@
+export const theme = {
+  borderRadius: {
+    xl: 'var(--f-borderRadius--xl)',
+    l: 'var(--f-borderRadius--l)',
+    m: 'var(--f-borderRadius--m)',
+    s: 'var(--f-borderRadius--s)',
+  },
+  breakPoint: {
+    sUp: '@media (min-width: 460px)',
```
ugogo, line 9: `sUp`? 🤔

src/components/FileUploader/index.js
```diff
@@ -181,14 +176,9 @@ FileUploader.propTypes = {
 
 FileUploader.defaultProps = {
   translate: ({ defaultText }) => defaultText,
-  hasError: false,
```
ugogo, line 184: Why did you remove all those props?

src/theme.js
obedparla: a bit of an awkward name 😄 . It's for `s` but up. Any better name comes to mind?

src/components/FileUploader/index.js
obedparla: `false` and `undefined` are both falsy, so they don't provide any value by being false, just extra code IMO. I checked that it doesn't change functionality

ugogo: 👌

src/theme.js
ugogo: Oh okay. Not really haha

PR #145, opened 2020-04-27
Obed/improvements v2

obedparla:
A few improvements before finishing v2

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@heetch/flamingo-react",
-  "version": "2.0.0",
+  "version": "1.9.3",
   "description": "Heetch Web Design System - Styled Components based",
   "author": "Heetch",
   "license": "MIT",
--- a/package.json
+++ b/package.json
@@ -11,6 +11,9 @@
   "engines": {
     "node": ">= 12.x"
   },
+  "browserslist": [
+    "defaults"
+  ],
   "precommit": "lerna run --concurrency 1 --stream precommit",
   "scripts": {
     "deploy": "lerna bootstrap --npm-client=yarn && lerna publish",
@@ -47,7 +50,7 @@
     "@svgr/rollup": "^4.3.3",
     "babel-eslint": "^10.1.0",
     "babel-loader": "^8.1.0",
-    "babel-plugin-styled-components": "^1.10.7",
+    "babel-plugin-styled-components": "1.10.5",
     "cross-env": "^6.0.3",
     "eslint": "^6.6.0",
     "eslint-config-airbnb": "^18.0.1",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@heetch/flamingo-react",
-  "version": "1.9.3",
+  "version": "2.0.0",
   "description": "Heetch Web Design System - Styled Components based",
   "author": "Heetch",
   "license": "MIT",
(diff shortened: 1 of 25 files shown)


src/components/Table/index.stories.js
```diff
@@ -8,7 +8,7 @@ import Icon from '../Icon';
 import Table from '.';
 import Text from '../Text';
 
-const data = [...new Array(50)].map(() => ({
+const data = [...new Array(7)].map(() => ({
```
tamarasaurus, line 11: That's a bit easier to read on the Storybook

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@heetch/flamingo-react",
-  "version": "1.9.3",
+  "version": "2.0.0",
```
ugogo, line 3: Nope nope nope, let Lerna handle that 🙂 
Also, we should release a RC before the v2

src/components/Alert/styles.js
```diff
@@ -3,10 +3,10 @@ import IconButton from '../IconButton';
 import { theme } from '../../theme';
 
 const styles = {
-  color({ type }) {
-    if (type === 'error') return theme.color.element.error;
-    if (type === 'success') return theme.color.element.success;
-    if (type === 'information') return theme.color.brand.secondary;
+  color({ alertType }) {
```
ugogo, line 6: Why did you rename this?

src/components/Table/index.js
```diff
@@ -20,12 +20,7 @@ const Table = React.forwardRef(
     },
     ref,
   ) => {
-    const {
-      headerGroups,
-      rows,
-      prepareRow,
-      state: { sortBy },
-    } = useTable(
+    const tableOps = useTable(
```
ugogo, line 23:
```suggestion
    const tableProps = useTable(
```
?

src/components/Alert/styles.js
obedparla: `type` is a valid html value so the dom was getting poluted

package.json
obedparla: ops. I was trying out if lerna worked. Will change

src/components/Table/index.stories.js
obedparla: I didn't follow, can you elaborate?

src/components/Alert/styles.js
ugogo: What about `intent` then? We already use this naming with the Button component

obedparla:
> Looks good 👍 , I just have another small suggestion. When I was using the Popover component I saw that even if the component is not visible (like `opacity:0`) it still overlays any elements beneath it, blocking clicks. I don't know if you already fixed that but it would be good to fix before v2.

Ahh, I didn't know this. I was thinking about changing how it animates. We can always fix it after V2

src/components/Alert/styles.js
obedparla: this is used only internally within `Alert`, the prop we use stays with the same name, I didn't want to change the name so much, just avoid it from going to the dom

src/components/Table/index.stories.js
tamarasaurus: I mean I don't wanna scroll down 50 rows to see the next example :smile:

obedparla: ah lol yeah 😄

PR #147, opened 2020-04-28
System font fix

keanan:
Removed the system font `BlinkMacSystemFont` which was not showing bolded header components in latest webkit browsers.

### DIFF
--- a/packages/css/src/shared/global.css
+++ b/packages/css/src/shared/global.css
@@ -3,8 +3,8 @@
 }
 
 body {
-  font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto,
-    Oxygen-Sans, Ubuntu, Cantarell, 'Helvetica Neue', sans-serif;
+  font-family: -apple-system, 'Segoe UI', Roboto, Oxygen-Sans, Ubuntu, Cantarell,
+    'Helvetica Neue', sans-serif;
   line-height: var(--f-lineHeight--m);
 }
 


obedparla:
why are you still working and how TF did you find this out.

obedparla:
Can you PR to `feature/styled-components` too? 🙏

keanan:
Hey @obedparla I just used dev tools to see what was causing the bold to not appear and knew it had something to do with system fonts as they are the only ones that should show up differently between browsers. 

In order to keep consistency across designs, we should not be using system fonts IMO.

Hey @ugogo this will only be fixed in version 83 and not 82, we are currently on 81:
https://phpnews.io/feeditem/chrome-vs-blinkmacsystemfont-when-bold-is-not-bold
https://www.coywolf.news/webmaster/chrome-81-breaks-system-fonts-bold/

Perhaps better to keep consistency until v83 is out?

obedparla:
> we should not be using system fonts IMO

Heartily agree. It can drastically change how our products look like for no real benefit

ugogo:
> we should not be using system fonts IMO
I agree but disagree, adding a webfont is terrible in a performance POV.
We could use system fonts, but more stable ones like Helvetica (OSX) and Arial (Windows)

keanan:
Hey @ugo yes indeed web fonts can be slow but there are ways to speed them up, found this good article:
https://medium.com/hackernoon/web-fonts-when-you-need-them-when-you-dont-a3b4b39fe0ae

Helvetica/Arial are a good choice, not OS dependent.
https://web.mit.edu/jmorzins/www/fonts.html

ugogo:
Let's rewrite everything in Papyrus :trollface:

keanan:
This has already been done in V2.

PR #148, opened 2020-04-29
Styled components/uniformize stories

ugogo:


ugogo:
https://github.com/heetch/flamingo/pull/149

PR #149, opened 2020-04-29
Styled components/uniformize stories 2

ugogo:
- Uniformize stories
- Update `yarn.lock`
- Fix `Card` shadows

### DIFF
--- a/src/components/Card/index.js
+++ b/src/components/Card/index.js
@@ -12,18 +12,10 @@ const styles = {
     return undefined;
   },
   boxShadow({ isSelected, elevation }) {
-    if (!isSelected && !elevation)
-      return `0 0 0 1px ${theme.color.element.inactive}`;
-
-    const shadows = [];
-
-    if (isSelected)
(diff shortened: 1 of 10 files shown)


src/components/Card/index.js
```diff
@@ -12,11 +12,18 @@ const styles = {
...
-    return `0 0 0 1px ${theme.color.element.inactive}`;
+    const shadows = [];
+
+    if (isSelected)
+      shadows.push(`inset 0 0 0 3px ${theme.color.brand.primary}`);
+
+    if (elevation === 2) shadows.push('0 4px 20px rgba(25, 1, 52, 0.14)');
+    if (elevation === 1) shadows.push('0 2px 10px rgba(25, 1, 52, 0.08)');
+
+    return shadows.join(', ');
```
obedparla, line 26: Why do we need to push shadows? The current implementation only has 1 shadow applied

ugogo: Oh I thought it was the case
Fixed

PR #155, opened 2020-05-05
Stronger image uploader validation

ugogo:
Some mobile devices don't support the `accept` prop on inputs, leading to app breaking because people tried to upload 2GB videos instead of a photo

### DIFF
--- a/src/components/ImageUploader/index.js
+++ b/src/components/ImageUploader/index.js
@@ -12,8 +12,20 @@ import { StyledImageUploader, UploaderPreview, PreviewHover } from './styles';
 const getDefaultValuePreviews = files =>
   Array.isArray(files) ? files.map(file => file.preview) : [];
 
+const validateFilesType = (files, accept) => {
+  const formats = accept.split(',');
+  return files.reduce((isValid, file) => {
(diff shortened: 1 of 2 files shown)


src/components/ImageUploader/index.js
```diff
@@ -12,8 +12,20 @@ import { StyledImageUploader, UploaderPreview, PreviewHover } from './styles';
 const getDefaultValuePreviews = files =>
   Array.isArray(files) ? files.map(file => file.preview) : [];
 
+const validateFilesType = (files, accept) => {
+  const formats = accept.split(',');
+  return files.reduce((isValid, file) => {
+    // if at least one file is not valid, return false
+    if (!isValid) return isValid;
+    return formats.includes(file.type);
```
groyoh, line 20: This may not work properly because:
1. `accept` may include spaces e.g. `accept="image/png, image/jpeg"`
2. `accept` may include wildcard e.g. `accept="image/*"`
3. `accept` may include extensions e.g. `accept=".png"`.

See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#Unique_file_type_specifiers.

PR #157, opened 2020-05-11
Init chakra-ui, add the theme foundations

ugogo:
Nothing crazy here
- Install `chakra-ui` and its dependencies
- Add the theme foundations (will evolve alongside new components)

NB: At the moment the theme and the provider are located into the storybook folder, it's temporary and will be moved later on

### DIFF
--- a/.storybook/theme/ThemeProvider.js
+++ b/.storybook/theme/ThemeProvider.js
@@ -1,0 +1,62 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {
+  CSSReset,
+  ThemeProvider as ChakraThemeProvider,
+  theme as chakraDefaultTheme,
+} from '@chakra-ui/core';
+
+// @TODO: move this into src/components
+// @TODO: speak with designers about the colors palette
+const theme = {
+  ...chakraDefaultTheme,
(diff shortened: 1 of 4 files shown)


.storybook/theme/ThemeProvider.js
```diff
@@ -0,0 +1,62 @@
...
+// @TODO: speak with designers about the colors palette
+const theme = {
+  ...chakraDefaultTheme,
+  breakpoints: ['28.75rem', '50rem', '68.75rem'],
+  colors: {
+    ...chakraDefaultTheme.colors,
+    primary: {
+      400: '#ff4d95',
+      500: '#ff2773',
+      600: '#ff4d95', // hover
```
obedparla, line 19: Is this how Chakra uses colors? Can't we keep it as we have it right now? `primary.whatever`

ugogo: Technically we can, but for example, Buttons will use `{color}.500` for background, `{color}.600` for hover, `{color}.400` for active, etc.
The idea is to have a wider palette with more explicit names, rather than `{color}-light|-light2` etc.
It's a point I need to talk to with the design team

PR #158, opened 2020-05-11
V3/feature branch

ugogo:
### ⚠️ This is a feature branch, please do not commit anything on it directly


src/components/Card/Card.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import { Box } from '@chakra-ui/core';
+import styled from '@emotion/styled';
+
+import { ITheme } from '../theme.d';
+
+export enum CardElevations {
```
amiiit, line 7:
```
type CardElevation = "none" | "s" | "m"
```

see for example this example: 
https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/Button/Button.d.ts#L18

src/components/Card/Card.tsx
```diff
@@ -0,0 +1,45 @@
...
+  NONE = '0',
+  S = '1',
+  M = '2',
+}
+
+export enum CardSizes {
+  S = 'S',
```
amiiit, line 14: same here, enum are not the way in this case but the example above

amiiit: also the use of plural is confusing IMO

src/components/Card/Card.tsx
ugogo: Both are valid options, 

Also, you can't iterate on `Type`s
https://github.com/heetch/flamingo/pull/158/files#diff-a78fa9cb2142c80dcfc7f47ee08fdaa3R32

amiiit: Just to save a line on storybook... I don't think it justifies using enum here, making the usage of the component too verbose.
If you really want, you could export a convenience const:
```
export CardElevations: CardElevation[] = ["none", "s","m"]
```

ugogo: It's not "just to save a line", it also ensures that a component's documentation is synced with its values
@heetch/frontend what do you guys think?

amiiit: I appologize if I underminded the usefulness of this. Either way, the solution suggested above should help you to cover that use case and a non-verbose usage.

obedparla: I agree it should be an `enum`, those are the only possible values it has, that's what an `enum` is for.

src/components/Icon/Icon.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from 'react';
+import * as Chakra from '@chakra-ui/core';
+
+export enum IconSize {
```
amiiit, line 4: 🌹   `type IconSize = 's' | 'm' | 'l'`, is a bit DRYer (:

src/components/Icon/Icon.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+export type IconProps = Chakra.IconProps & {
+  size?: IconSize | string;
+};
+
+const getSize = (size: string) => {
+  if (size === 's') return '0.75rem';
+  if (size === 'm') return '1rem';
+  if (size === 'l') return '1.25rem';
+  return size;
```
amiiit, line 18: otherwise return a default size 1em?

ugogo: There is a default size applied in props `size = IconSize.m`

src/components/Icon/Icon.tsx
ugogo: Agree, but if we want to stay consistent with the other components, it should be an `enum`

amiiit: I absolutely respect this argument. But I personally believe that consistency shouldn't stand in the way to better architecture as long as it's properly documented (we're using TS, so it is documented), as it's all still dynamically evolving.
That is just a suggestion ...

ugogo:
:rip:

PR #160, opened 2020-05-12
V3/migrate card

ugogo:
# Migrate `Card` component

## Breaking changes
- The `isSeleted` prop has been renamed to `isActive`
- Elevations enum changed to
```ts
export enum Elevations {
  NONE = '0',
  S = '1',
  M = '2',
}
```

## Changes to discuss with designers
- Are the nested styles needed?
**➡️ yes**

## PR description
Ok, quite a lot is happening here 😄 
I added some comments trying to clarify the changes


.eslintrc
```diff
@@ -35,6 +35,21 @@
...
+    "no-useless-constructor": 0,
+    "@typescript-eslint/no-useless-constructor": "error",
+    "react/prop-types": 0
+  },
+  "settings": {
+    "import/resolver": {
+      "node": {
+        "extensions": [".js", ".jsx", ".ts", ".tsx"]
+      }
+    }
```
ugogo, line 53: Needed to not having an error when importing `.ts/.tsx`

.storybook/theme/ThemeProvider.js
```diff
@@ -39,6 +59,15 @@ const theme = {
...
     xl: '1.5rem',
   },
+  shadows: {
+    ...chakraDefaultTheme.shadows,
+    Card: {
+      isActive: `inset 0 0 0 3px ${colors.primary[500]}`,
+      0: `0 0 0 1px ${colors.element.inactive}`,
+      1: '0 2px 10px rgba(25, 1, 52, 0.08)',
+      2: '0 4px 20px rgba(25, 1, 52, 0.14)',
+    },
```
ugogo, line 69: I moved this here, thinking that it would be nice to have this kind of stuff centralized in the theme
Open to discussions 🙂

src/components/Card/Card.d.ts
```diff
@@ -0,0 +1,16 @@
...
+}
+
+export enum Sizes {
+  S = 'S',
+  M = 'M',
+}
+
+export interface IProps {
+  elevation?: Elevations;
+  isActive?: boolean;
```
ugogo, line 14: Previously `isSelected`, WDYT?

src/components/Card/Card.tsx
```diff
@@ -0,0 +1,29 @@
...
+  <Box
+    bg='element.primary'
+    p={size === Sizes.S ? 'xl' : 'xxl'}
+    shadow={isActive ? 'Card.isActive' : `Card.${elevation}`}
+    rounded='xl'
+    {...props}
+  />
+);
+
+export default styled(Card)`
```
ugogo, line 23: `styled` is needed here because of the use of `&` for nested instances

src/components/theme.d.ts
```diff
@@ -0,0 +1,9 @@
+import { ITheme as ChakraITheme } from '@chakra-ui/core';
```
ugogo, line 1: Will evolve with the refactoring, will then be into the `ThemeProvider` folder (or `theme` one?)

src/components/Card/Card.d.ts
```diff
@@ -0,0 +1,16 @@
...
+  S = '1',
+  M = '2',
+}
+
+export enum Sizes {
+  S = 'S',
+  M = 'M',
+}
+
+export interface IProps {
```
obedparla, line 12: A few things:

1) is it the standard to name the interfaces/TS stuff as `<file>.d.ts`? It would be good to follow whatever is most common
2) I didn't see this being exported to the main bundle, or through the main component. Maybe I missed it?
3) We should decide from now how we're naming interfaces. I think we can have better names than `I<Props>`, such as simply `Card` or `ICard`

tamarasaurus: `I<Props>` is pretty common to name interfaces in TS, but I prefer the naming style of `CardProps` `CardState` etc..

tamarasaurus:
Will the types be exported into a single d.ts file in the end?

src/components/Card/Card.d.ts
ugogo: > is it the standard to name the interfaces/TS stuff as <file>.d.ts? It would be good to follow whatever is most common

I think so, I recall seeing this into multiple projects

> I didn't see this being exported to the main bundle, or through the main component. Maybe I missed it?

The build is actually failing because of `[!] Error: Could not resolve './Card.d' from src/components/Card/Card.tsx`, but I don't understand why :cry:

> I<Props> is pretty common to name interfaces in TS, but I prefer the naming style of CardProps CardState etc...

Looks good to me, should we do the same for enums? `CardElevations`, `CardSizes`?

> Will the types be exported into a single d.ts file in the end?

Once the build is fixed, we can created a `index.d.ts` file exporting every types

obedparla: +1 to naming things Card<Something>

ugogo: Renaming ✅ 
I was able to fix the build by moving the type definitions into the component file, but IDK if that's a good practice. WDYT?

ugogo: 🆙 @obedparla @tamarasaurus

rollup.config.js
```diff
@@ -28,7 +28,12 @@ export default {
     babel({
       babelHelpers: 'bundled',
     }),
-    commonjs(),
+    commonjs({
+      namedExports: {
+        'body-scroll-lock': ['disableBodyScroll', 'enableBodyScroll'],
```
keanan, line 33: Nice, can use this for the side panel when it is open

src/components/Card/Card.d.ts
ugogo: Types are correctly exported, let's keep types into the component file

PR #164, opened 2020-05-25
V3/migrate heading

ugogo:
# Migrate `Heading` component

Since components won't come with vertical spacing anymore, I created a `StoryHeading` one.
Most of the changes are related to this
If you want quick access to the refactored Heading component, you can click here: https://github.com/heetch/flamingo/pull/164/files#diff-851f5ae413bc8a9023d063b6c2976c2a

## Breaking changes
- None


.storybook/theme/ThemeProvider.js
```diff
@@ -53,6 +53,12 @@ const theme = {
     xl: '1.5rem',
     xxl: '2rem',
   },
+  lineHeights: {
+    s: '1rem',
+    m: '1.4rem',
+    l: '1.5rem',
```
keanan, line 59: How come 1.4 for M and 1.5 for L - should it not be somewhere closer to the center like 1.25 for M?

ugogo: I followed the actual values from `styles.css`
https://github.com/heetch/flamingo/blob/master/src/styles.css#L44-L47

PR #165, opened 2020-05-29
Add defaultIndex to Tabs

groyoh:
This defaultIndex allows us to define which tab should be open on initial render.

### DIFF
--- a/src/components/Tabs/index.js
+++ b/src/components/Tabs/index.js
@@ -9,10 +9,20 @@ import {
 } from './styles';
 
 const Tabs = React.forwardRef(
-  ({ elements, buildTabLabel, buildReturnObject, onClick, ...props }, ref) => {
-    const [activeIndex, setActiveIndex] = useState(null);
+  (
+    {
+      elements,
+      defaultIndex,
+      buildTabLabel,
+      buildReturnObject,
+      onClick,
+      ...props
+    },
+    ref,
+  ) => {
+    const [activeIndex, setActiveIndex] = useState(defaultIndex);
 
-    const handleTabClick = newActiveIndex => {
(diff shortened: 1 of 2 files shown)


src/components/Tabs/index.stories.js
```diff
@@ -11,6 +11,7 @@ stories.add('Playground', () => (
   <>
     <Heading>Tabs</Heading>
     <Tabs
+      defaultIndex={1}
```
ugogo, line 14: Shouldn't it be `0`?

src/components/Tabs/index.js
```diff
@@ -9,10 +9,20 @@ import {
...
+      buildReturnObject,
+      onClick,
+      ...props
+    },
+    ref,
+  ) => {
+    const [activeIndex, setActiveIndex] = useState(defaultIndex);
 
-    const handleTabClick = newActiveIndex => {
+    const handleTabClick = (newActiveIndex) => {
```
keanan, line 25: don't need these?

src/components/Tabs/index.stories.js
groyoh: Well, it's to showcase that you can select any index. So it can be anything.

src/components/Tabs/index.js
groyoh: Looks like my prettier failed locally.

src/components/Tabs/index.stories.js
ugogo: My bad I thought it was done on the component side 🤦 
In that case, could you create a new section to showcase this?

```
<Heading>Tabs</Heading>
...
<Heading level={2}>With default index</Heading>
    <Tabs defaultIndex={1} ... />

```

PR #171, opened 2020-06-17
Merge uploaders component

ugogo:
This PR aims to merge both `FileUploader` and `ImageUploader` components.
They both serve the same purpose, with a little tweak to display a single image.

### Changes
We are now relying on a `state` to control the styling and what's rendered by the component, rather than a bunch of constants and props.
https://github.com/heetch/flamingo/blob/25682a3abe114f229c8b2c461b0f357b3b266210/src/components/Uploader/constants.js

I also updated the stories to have a larger representation of what's the component can render
https://flamingo-nnlbufcpk.vercel.app/?path=/story/form-uploader--all-states

### Behavior
When a single file is uploaded, and it's an image -> display a large preview
Otherwise, display a list of `UploaderItem`.
A check is done to know if the file passed into the `UploaderItem` is an image or not.
If it is, it will load a preview


src/components/Uploader/index.js
```diff
@@ -0,0 +1,222 @@
...
+
+      setPreview(await toBase64(inputFile));
+      setState(STATES.SINGLE_IMAGE);
+    };
+
+    const onFilesChange = inputFiles => {
+      setState(STATES.LOADING);
+      onChange(inputFiles);
+
+      if (inputFiles.length === 1) onSingleFileChange(inputFiles[0]);
```
obedparla, line 97: Instead of going into singleFileChange to check if it's an image just to execute multipleFiles again, why not do it here directly instead? 

```suggestion
      if (inputFiles.length === 1 && isImage(inputFile)) onSingleFileChange(inputFiles[0]);
```

src/components/_DeprecatedFileUploader/index.js
```diff
@@ -67,6 +67,11 @@ const FileUploader = React.forwardRef(
 
     // keep files in sync with value when changed on image uploader
     React.useEffect(() => {
+      // eslint-disable-next-line no-console
+      console.warn(
+        '[@heetch/flamingo] The FileUploader component has been deprecated. Please use the Uploader one',
```
obedparla, line 72: 👍

src/components/Uploader/index.js
```diff
@@ -0,0 +1,222 @@
...
+    id: 'flamingo.file-uploader.upload-more-files',
+    defaultText: 'Add more files',
+  },
+  uploading: {
+    id: 'flamingo.file-uploader.uploading',
+    defaultText: 'Uploading...',
+  },
+};
+
+const isImage = file => /\.(jpe?g|png|gif)$/i.test(file.name);
```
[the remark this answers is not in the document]
ugogo: I don't think so 😄 
https://caniuse.com/#feat=webp

src/components/Uploader/index.js
ugogo: Good point 👍

ugogo: Done in https://github.com/heetch/flamingo/pull/171/commits/582d8e84c22391e590dbcbe9f7f2083d71aa01e1
WDYT @obedparla?

ugogo: Nevermind, I merged both fn into a single one, it's even better 😄 
https://github.com/heetch/flamingo/pull/171/commits/4a93e74ca25e18f2fb2ce24576cdf1b101cb07a4

obedparla: Looks great 👍

src/components/Uploader/index.js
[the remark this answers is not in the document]
obedparla: but what happens if a user uploads a webp image though? it will treat it as a file. Most users won't know what type of image they're using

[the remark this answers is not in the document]
ugogo: Ok I tried with a `.webp` file and it converts as expected.
But even macOS doesn't support it, the Prevew app can't load it 🤷

PR #172, opened 2020-06-17
Rework Text variant prop

ugogo:
Dropping the `variant` prop of the `Text` component will bring a more granular approach on how we build our components

Where `<Text variant={Text.variants.SUBCONTENTBOLD />` wasn't really clear (also, why uppercase?), `<Text size="small" weight="bold" />` is easier to read and understand what we're doing

NB: we can still use `<Text size={Text.TextSizes.small} weight={Text.TextWeights.bold} />` if needed


src/components/Text/Text.tsx
```diff
@@ -1,37 +1,31 @@
 import * as React from 'react';
 import { Text as ChakraText, BoxProps } from '@chakra-ui/core';
 
-export enum TextVariants {
-  CONTENT = 'CONTENT',
-  CONTENTBOLD = 'CONTENTBOLD',
-  SUBCONTENT = 'SUBCONTENT',
-  SUBCONTENTBOLD = 'SUBCONTENTBOLD',
+export enum TextWeights {
+  normal = 'normal',
```
obedparla, line 5: I'd add italic here too.

src/components/Text/Text.tsx
```diff
@@ -1,37 +1,31 @@
...
-};
-
-const fontSizeByVariant = {
-  [TextVariants.CONTENT]: 'm',
-  [TextVariants.CONTENTBOLD]: 'm',
-  [TextVariants.SUBCONTENT]: 's',
-  [TextVariants.SUBCONTENTBOLD]: 's',
-};
+export enum TextSizes {
+  normal = 'normal',
```
obedparla, line 10: What about simplifying this and using a boolen for `bold`? Like `<Text bold />`. If bold is not present then it's normal. I used to have that  in my previous design system and it was nice

ugogo: Mh 🤔 
What if we add more weights later on?
Would you do the same for the size?

This is something we don't do on the other components (like in `Card` for example, we do `size="small"` and not `small`

IMO it's safer and more clear to go for an enum, rather than a prop
How would you expect the component to behave if we do something like `<Text bold light />`?

Any opinion @heetch/frontend?

tamarasaurus: I prefer enums for that, I wouldn't want to define a new prop every time we have a new weight

ugogo: Works for you @obedparla?

src/components/Text/Text.tsx
ugogo: It's not a weight, and not part of the specs IIRC

src/components/Text/Text.tsx
obedparla: sounds good to me, I wanted to pur the idea out there 👍

src/components/Text/Text.tsx
obedparla: Right 🤦 I always forget that haha

PR #177, opened 2020-06-19
Fix props spreading in form fields

ugogo:


obedparla:
What's the reason for this change? We'd need to debug all form elements in the apps to make sure this doesn't create a bug

ugogo:
@obedparla My bad, I made a wrong assumption 🤦 
Props were leaking on the `Uploader` container, but the issue on the `Uploader` itself and not on the `FormField` 😬

PR #179, opened 2020-07-01
V3/migrate icons

ugogo:
# Migrate `Icon` component

## Breaking changes
- Icon sizes and list are not available anymore via the component (`Icon.ICONS.x`, `Icon.SIZES.x`)

## Improvements
- The `size` prop now accept a custom value, in addition to the current `s|m|l` ones

## Notes
- The previous `Icon` component has been renamed `_DeprecatedIcon` to not break other components implementation


.storybook/theme/icons.js
```diff
@@ -0,0 +1,1309 @@
+import * as React from 'react';
```
ugogo, line 1: This will be moved later on with the `ThemeProvider` component

scripts/icons.js
```diff
@@ -88,10 +82,16 @@ const formatIcons = iconsPath.map(async iconPath => {
...
+  const output = `import * as React from 'react';
+
+/*
+ * This file has been automatically generated.
+ * Do not update it manually, run 'yarn icons' instead.
+ */
 
-/* This file has been generated via the 'icons' script */
-${reactIcons}`;
+export default {
```
ugogo, line 92: Icons are not in an object, rather than in a React component

src/components/Icon/Icon.stories.tsx
```diff
@@ -0,0 +1,65 @@
...
+import Icon, { IconSize } from './Icon';
+import Input from '../Input';
+import Label from '../Label';
+
+type IconsFiltererProps = {
+  children: (icons: string[]) => React.ReactNode;
+  filter?: string;
+};
+
+const IconsLister = ({ children, filter = '' }: IconsFiltererProps) => {
```
ugogo, line 18: For some reasons, the `useTheme` hook wasn't returning the theme inside a story 🤷

obedparla: is the story wrapped in a ThemeProvider?

ugogo:
> Why are all the icons imported from _DeprecatedIcon instead of the new Icon, is there some limitation for now to do it that way, or it's just to avoid changing a bunch of other components?

Because I didn't want to introduce too many changes
The DeprecatedIcon component will be removed when refactoring the wrapping components

src/components/Icon/Icon.stories.tsx
ugogo: Yes, cf: https://github.com/heetch/flamingo/pull/158/files#diff-41e445b678933963d77632d4206c90b8R17
It doesn't work only when the hook is used at the top level 🤷

PR #180, opened 2020-07-03
Make component propTypes SSR friendly

tamarasaurus:
By adding a small check for proptypes that refer to `HTMLElement` or `Element` these components would be safe to use with SSR

### DIFF
--- a/src/components/InputField/index.js
+++ b/src/components/InputField/index.js
@@ -17,7 +17,9 @@ InputField.propTypes = {
   id: PropTypes.string.isRequired,
   label: PropTypes.node,
   inputRef: PropTypes.shape({
-    current: PropTypes.instanceOf(Element),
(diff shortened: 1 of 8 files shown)


src/components/InputField/index.js
```diff
@@ -17,7 +17,7 @@ InputField.propTypes = {
   id: PropTypes.string.isRequired,
   label: PropTypes.node,
   inputRef: PropTypes.shape({
-    current: PropTypes.instanceOf(Element),
+    current: typeof Element !== 'undefined' ? Element : Object,
```
obedparla, line 20: shouldn't it contain `PropTypes.instanceOf`?

obedparla: same for the other ones

tamarasaurus: Yeah :facepalm: fixed!

PR #182, opened 2020-07-09
V3/convert enums

ugogo:
Context: https://heetch.slack.com/archives/C2BHK4Q4Q/p1594281037010900
`as const` are needed, see https://stackoverflow.com/questions/37978528/typescript-type-string-is-not-assignable-to-type

### DIFF
--- a/src/components/Card/Card.stories.tsx
+++ b/src/components/Card/Card.stories.tsx
@@ -3,11 +3,14 @@ import { Box } from '@chakra-ui/core';
 import { storiesOf } from '@storybook/react';
 import { boolean, select } from '@storybook/addon-knobs';
 
-import Card, { CardElevations, CardSizes } from './Card';
+import Card from './Card';
 import StoryHeading from '../StoryHeading/StoryHeading';
 import Text from '../Text/Text';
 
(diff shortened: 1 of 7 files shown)


src/components/Card/Card.stories.tsx
```diff
@@ -3,11 +3,14 @@ import { Box } from '@chakra-ui/core';
...
 
-import Card, { CardElevations, CardSizes } from './Card';
+import Card from './Card';
 import StoryHeading from '../StoryHeading/StoryHeading';
 import Text from '../Text/Text';
 
-const Container = props => <Box m='xl' {...props} />;
+const Container = (props: any) => <Box m='xl' {...props} />;
+
+const elevations = ['none', 's', 'm'] as const;
```
amiiit, line 12: Out of interest: what `as const` means?

ugogo: As explained in the PR description: https://stackoverflow.com/questions/37978528/typescript-type-string-is-not-assignable-to-type

It's basically a cast preventing TS to throw an error

amiiit: Interesting... thanks for the info

PR #183, opened 2020-07-10
V3/uniformize imports

ugogo:
Uniformize imports

```diff
-import React from 'react';
+import * as React from 'react';
```

```diff
-import { ... } from '@chakra-ui/core';
+import * as Chakra from '@chakra-ui/core';
```

### DIFF
--- a/.storybook/theme/StoryWrapper.js
+++ b/.storybook/theme/StoryWrapper.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import * as React from 'react';
 
 export default storyFn => (
   <div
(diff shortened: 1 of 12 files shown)


.storybook/theme/StoryWrapper.js
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
+import * as React from 'react';
```
obedparla, line 1: I read somewhere this yields different results behind the curtain but I haven't learned about it, do you know what the difference is? 🤔

ugogo: IIRC this is the recommended way of importing React

From https://github.com/facebook/react/pull/18102
```
This is the correct way to import React from an ES module since the ES module will not have a default export. Only named exports.
```

PR #186, opened 2020-07-31
Section and pagination components

obedparla:
- Migrate the Section and pagination components from the admin to Flamingo 
- Add support for Jest, since the Pagination comp already had tests I wanted to include them


src/components/Pagination/index.js
```diff
@@ -0,0 +1,85 @@
...
+
+  const { translations } = props;
+
+  return (
+    <Container className={props.className} position={props.position}>
+      <ArrowIcon
+        data-testid={'pagination-prev'}
+        icon={'IconArrowLeft'}
+        onClick={props.goPrev}
+        css={{ marginLeft: '24px' }}
```
ugogo, line 22: `--f-space--xxl`

src/components/Pagination/index.js
```diff
@@ -0,0 +1,85 @@
...
+import IconButton from '../IconButton';
+import { theme } from '../../theme';
+
+const Pagination = props => {
+  // Don't show it if both are disabled
+  if (props.prevDisabled && props.nextDisabled) {
+    return null;
+  }
+
+  const { translations } = props;
```
ugogo, line 14: Please use the same pattern as in the Uploader component
https://github.com/heetch/flamingo/blob/master/src/components/Uploader/index.js#L43

src/components/Pagination/index.js
```diff
@@ -0,0 +1,85 @@
...
+        onClick={props.goPrev}
+        css={{ marginLeft: '24px' }}
+        disabled={props.prevDisabled}
+        title={`${translations.previous} ${
+          props.prevDisabled ? `- ${translations.disabled}` : ''
+        }`}
+      />
+
+      <ArrowIcon
+        data-testid={'pagination-next'}
```
ugogo, line 30: Is it still a good practice?
I remember KCD saying that it should be the last option to use 🤔

src/components/Pagination/index.stories.js
```diff
@@ -0,0 +1,28 @@
...
+stories.add('Playground', () => {
+  const [page, setPage] = useState(0);
+  return (
+    <>
+      <Heading>Pagination</Heading>
+      <Heading level={2}>Page #{page}</Heading>
+
+      <Pagination
+        prevDisabled={page === 0}
+        nextDisabled={page === 3}
```
ugogo, line 21: How would you implement that in a real-life situation?

src/components/Section/index.js
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import { theme } from '../../theme';
+import UiText from '../UiText';
+
+const Section = props => (
```
ugogo, line 7: Destructuring 😄

src/components/Section/index.js
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
```
ugogo, line 1: I think this component should be part of the Card one

src/components/Section/index.stories.js
```diff
@@ -0,0 +1,45 @@
...
+const defaultProps = () => ({
+  id: `section-${Math.random()}`,
+  title: text('Title', 'Title'),
+  subtitle: text('Subtitle', 'Subtitle'),
+  children: text('Children', 'How you doin'),
+});
+
+const stories = storiesOf('Section', module);
+
+stories.add('All states', () => (
```
ugogo, line 17: Nice 👌

src/components/Pagination/index.js
obedparla: Yeah, but there is no other way to select it since it's a icon

src/components/Pagination/index.stories.js
obedparla: You can check the `hashtag management` page or the `vouchers` in the admin for its usage. 

But basically it depends on your factors. Sometimes the logic comes from the values in a data object, sometimes from something else.  The component doesn't know what page it's on, hence the need for it

src/components/Pagination/index.js
```diff
@@ -0,0 +1,85 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+
+import IconButton from '../IconButton';
+import { theme } from '../../theme';
+
+const Pagination = props => {
```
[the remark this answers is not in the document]
obedparla: Arg 😄  we're not so strict in the admin with that and I copy pasted it

src/components/Section/index.js
obedparla: I don't want to mix them tho. They serve different purposes, like you can have many cards next to each other, but a section is just a ... section 😄 . But I extended the container with `Card` so it can have the same `box-shadow` props tho. Maybe it'll come handy.

ugogo: 👌

src/components/Pagination/index.stories.js
ugogo: I see 🤔 
Wouldn't it be simpler to pass a `currentPage` and a `totalPage` sort of props?

src/components/Pagination/index.js
ugogo: 🆙 
`translate` is a function returning an `id` and a fallback
The goal would be to use it as `translate={t}` to translate the whole component, where `t` can be pulled from `i18next` for example

src/components/Pagination/index.stories.js
obedparla: indeed, but we don't always know how many pages there are tho.

obedparla: for example in `user search` the logic for pagination is special. We don't know how many pages there are but each page has 100 items. So the `nextDisabled === { if current page has less than 100 items}`

PR #187, opened 2020-09-17
Add pagination table component

keanan:
### :blue_book: What does this PR do and why is this needed?

Add new paginated table component

### :pencil: Describe the important code changes

Add new paginated table component
Add new colour var `lightGrey` for table row hover

### :grey_question: Questions or remarks

The paginated table is very similar to the regular one we currently use with the exception of the React Table using pagination hooks and some extra props `pageSize` and `pageIndex`. 

3 new custom props have been added which we currently use: `onRowClick`, `colWidth` and `disableColSort` (@tamarasaurus @OctaveStoufflet I can also add these to the regular `Table` wanted to show you first - since the `min-width` has been removed from the column header styles).

`disableColSort` does not completely disable the sort function from React Table as it applies to the entire table - the best solution I could find was to at least remove the arrows, hover cursor and change in text color - if anyone has any suggestions how to improve this please share, thanks.


OctaveStoufflet (approved):
All good for me ! 🔥
Don't you think, it could be nice to also add the meta informations ? (3/10)
As we need to display it when our tables has several pages... 🌷

keanan:
Hey @OctaveStoufflet, great idea! currently the meta information is included from the Pagination component manually, so I will make a note add it to the Pagination component as a prop we can use if desired, thanks.

PR #190, opened 2020-10-21
Allow onClick event for table rows

tamarasaurus:
This PR adds the prop `onClickRow` to the `Table` component. When a row is clicked, the original data values will be passed to the handler. E.g.

```javascript
const data = [
  { name: 'Tam', active: true },
  { name: 'Paco', active: false },
];

return (
  <Table
    data={data}
    columns={[
      {
        Header: 'Name',
        accessor: 'name',
      },
      {
        Header: 'Name',
        accessor: 'name',
      },
    ]}
    onClickRow={row => console.log(row)} //outputs { name: X, active: X }
  />
);

```

### DIFF
--- a/src/components/Table/index.js
+++ b/src/components/Table/index.js
@@ -17,6 +17,7 @@ const Table = React.forwardRef(
       manualSorting,
       onChangeSort,
       initialState,
+      onClickRow,
     },
     ref,
   ) => {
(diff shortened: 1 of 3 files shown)


OctaveStoufflet (approved):
Thank you Tam perfect 👍

keanan (approved):
Nice, thanks for doing this

PR #191, opened 2020-11-04
Switch to react-datepicker 

tamarasaurus:
This PR removes the existing `Datepicker` component (using react-day-picker) and adds `DatepickerDay` (using react-datepicker). The new library allows us to have manual input and validation of dates, and easily pass in locale data from `date-fns`. For now I've only done the standard calendar picker for a day, but we could incrementally add the other components like `DatepickerMonth`, `DatepickerRange` etc.. 

Screenshot from 2020-11-05 15-08-55

To pass in a locale:
```javascript
import fr from 'date-fns/locale/fr';

...

<DatepickerDay
    locale={{ key: 'fr', dateFnsLocale: fr }}
    placeholder='Choisir une date' // The placeholder must be manually translated
    onChange={() => {}}
/>
```

I also made another change to the way rollup/postcss includes third party CSS from node_modules. Instead of importing the react-datepicker styles in JS I moved the import statement to CSS. This won't make much of a different except that those styles will no longer be injected into the page at runtime (but in the dist/styles.css instead), and we won't see them in the output of our unit tests anymore :tada: 

Design QA passed :heavy_check_mark:


keanan (approved):
This is 😍  Thanks for doing this 🙌

PR #192, opened 2020-12-02
Update Table proptypes for accessor

tamarasaurus:
This PR updates the prop type definitions for the Table component to remove the error messages when we have a valid definition of `accessor` as a function

### DIFF
--- a/src/components/Table/index.js
+++ b/src/components/Table/index.js
@@ -107,7 +107,7 @@ Table.propTypes = {
   columns: PropTypes.arrayOf(
     PropTypes.shape({
       Header: PropTypes.string,
-      accessor: PropTypes.string,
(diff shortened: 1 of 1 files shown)


src/components/Table/index.js
```diff
@@ -107,7 +107,7 @@ Table.propTypes = {
   columns: PropTypes.arrayOf(
     PropTypes.shape({
       Header: PropTypes.string,
-      accessor: PropTypes.string,
+      accessor: PropTypes.oneOfType([PropTypes.func, PropTypes.string]),
```
OctaveStoufflet, line 110: let me put any type of accessor bitch !

OctaveStoufflet (approved):
thanks Tam !!

PR #201, opened 2021-04-14
[TCC-1410] [FE] Add Badge component to Flamingo

OctaveStoufflet:
This PR add our Badge in the flamingo library

### DIFF
--- a/src/components/Badge/styles.js
+++ b/src/components/Badge/styles.js
@@ -1,0 +1,13 @@
+import styled from 'styled-components';
+import { theme } from '../../theme';
+
+export const RoundedBadge = styled.div`
+  display: inline-block;
+  border-radius: ${({ borderRadius }) => borderRadius || theme.borderRadius.xl};
+  padding: 2px 10px;
+  color: ${({ color }) => color || theme.color.text.primary};
+  background-color: ${({ backgroundColor }) =>
+    backgroundColor || theme.color.element.primary};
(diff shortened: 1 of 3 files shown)


src/components/Badge/styles.js
```diff
@@ -0,0 +1,9 @@
+import styled from 'styled-components';
+import { theme } from '../../theme';
+
+export const RoundedBadge = styled.div`
```
ganlhi, line 4: It should be an inline block instead of a block, so either:
- make it a `styled.span`
- add `display: inline-block` to it

OctaveStoufflet: It’s a draft :) I’ve added the wip tag. No need to review this PR for now, thanks !

ganlhi: My bad 🙊

ganlhi (approved):

src/components/Badge/styles.js
```diff
@@ -0,0 +1,13 @@
...
+  display: inline-block;
+  border-radius: ${theme.borderRadius.xl};
```
tamarasaurus, line 6: Do you plan to make the border radius a prop on the component?

OctaveStoufflet: No.. but I can ! Indeed we talked about that :load:

tamarasaurus (approved):

PR #206, opened 2021-04-28
update react-popper to 2.2.5

OctaveStoufflet:
This PR update Popper to its last version, and add more props to our Popper component to be more flexible.

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,10 +94,11 @@
     "dist"
   ],
   "dependencies": {
+    "@popperjs/core": "^2.9.2",
     "classnames": "^2.2.6",
     "normalize.css": "^8.0.1",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -94,10 +94,11 @@
     "dist"
   ],
   "dependencies": {
+    "@popperjs/core": "^2.9.2",
```
tamarasaurus, line 97: For the `Dropdown` component, comparing the old and the new Storybooks it looks like the 'click outside to close' behaviour doesn't work anymore and the spacing is not the same between the 'triggerer' and the dropdown lists.

ganlhi (approved):

tamarasaurus (approved):

PR #221, opened 2021-09-14
[DAB] Flamingo 3.0

agathepey:
We are updating Driver Portal with a new design (aligned with the design flamingo 3.0 we have on the mobile app).

- Most of components have new designs
- Changed styled-components to macro
- This version of flamingo will be known as alpha version 
- README is updated, to use this alpha version

Here is the [Figma](https://www.figma.com/file/c8swDK3vAYXoINOECXJsSd/Driver-Portal-3.0-Migration?node-id=200%3A9896)

Note : Driver Portal PR with this new version is [here](https://github.com/heetch/universe/pull/15079)


ganlhi:
Hey @agathepey! Could you share the specs this work is based on?

agathepey:
> Hey @agathepey! Could you share the specs this work is based on?

Sure 😃   Just to give you a bit of context (because we discussed about it during one of our FE meeting, but I'm not sure you were here):
We are creating a new flamingo version to have our design on Driver Portal based on the same design we have on the mobile app.
[Here is the Figma](https://www.figma.com/file/c8swDK3vAYXoINOECXJsSd/Driver-Portal-3.0-Migration?node-id=200%3A9896)

Btw, (because I know you can be worried about it 😅) we will use only this version on Driver Portal for now. Don't hesitate if you have questions 🤗

ganlhi:
Thanks for the details! I want to keep an eye on it, because even if it's only for the driver portal for now, I would prefer we don't lag behind for too long :)

src/components/Toast/index.stories.js
```diff
@@ -23,11 +23,13 @@ stories.add('Playground', () => {
         position={'top'}
         onClose={onClose}
       >
-        {"I'm a subtitle"}
+        {"I'm a subtitle with a "}
+        <a href='https://example.com'>link</a>
```
ganlhi, line 27: In this case it's not critical, but be sure to update the `<Code />` examples when you update a story's code

agathepey: Ah good catch, forgot about that, thanks 🤗

src/components/Button/styles.js
```diff
@@ -31,39 +32,45 @@ const styles = {
     return theme.color.text.white;
   },
   backgroundColor({ intent, disabled, variant }) {
-    if (disabled) return `${theme.color.element.primary} !important`;
+    if (disabled) return `${theme.color.element.inactive} !important`;
```
ganlhi, line 35: image

I think the text variant should have the text color grayed out when disabled instead of looking like the normal button disabled.

src/components/Button/styles.js
```diff
@@ -139,15 +139,13 @@ export const StyledButton = styled('button').attrs(() => ({
...
   }
-  ${({ variant }) =>
-    variant !== 'text' &&
-    `
+
   &:hover {
     background-color: ${styles.hover.backgroundColor};
-    color: ${theme.color.text.white};
+    color: ${({ variant }) =>
+      variant === 'text' ? styles.hover.color : theme.color.text.white};
```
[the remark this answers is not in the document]
ganlhi:
```suggestion
      variant === 'text' ? undefined : theme.color.text.white};
```
This is enough to have the color remain the same on hover.

src/theme.js
```diff
@@ -154,8 +154,8 @@ export const theme = {
       },
     },
     bg: {
-      primary: 'var(--f-color-ui-default)',
-      secondary: 'var(--f-color-ui-secondary)',
+      primary: 'var(--f-color-ui-secondary)',
```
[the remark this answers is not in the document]
ganlhi: I just checked, `--f-color-ui-secondary` doesn't exist, so definitely this needs to be `--f-color-bg-primary`

ganlhi:
Last commit is about having textareas as in https://www.figma.com/file/knV0XSSs86zy6VAiCPRAEh/FAQ?node-id=395%3A7071

ganlhi (approved):

ganlhi:
Now it is rebased on master (with Node 16), we are ready to merge, then release 5.1.0.

keanan (approved):
Big effort from us all, thank-you! 👏

PR #226, opened 2022-01-06
Update react-scripts

keanan:
Update react-scripts to bump `immer` dependency

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
     "prettier": "^1.19.1",
     "react": "^16.8.6",
     "react-dom": "^16.8.6",
-    "react-scripts": "^3.4.1",
+    "react-scripts": "4.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -76,7 +76,7 @@
     "prettier": "^1.19.1",
     "react": "^16.8.6",
     "react-dom": "^16.8.6",
-    "react-scripts": "^3.4.1",
+    "react-scripts": "4.0.0",
```
ganlhi, line 79: Why not `^4.0.0`? Is there a particular reason?

keanan: I wanted to see which package dependencies would get updated and bumped it according to the docs on migration from 3.4.4 -> 4.0.0

This is a `wip` and wanted to take some time to discuss options in our meeting tomorrow.

ganlhi: Works for me :)

PR #235, opened 2022-04-27
[TCC-2443] flamingo-react Necessary updates

OctaveStoufflet:


src/components/Button/index.stories.js
```diff
@@ -12,7 +12,7 @@ import { theme } from '../../theme';
 const { INTENTS, VARIANTS } = Button;
 
 const intents = ['primary', 'secondary'];
-const variants = ['outline', 'minimal'];
+const variants = ['outline', 'minimal', 'text'];
```
ganlhi, line 15: I think there is something missing for this variant: the [hover](https://github.com/heetch/flamingo/pull/235/files#diff-50891f9556ef9241803a61a0e27339601b097c50487f47325da08e3d659b7d18R69) should not change the text color.

PR #246, opened 2022-09-27
Uploader - Icon always visible

agathepey:
Linked to [DAB-1169] & Pierre Lange's comment. 

Having the `IconButton` with an unset color on the Uploader is not good UX. 
We don't know that we have to go to the top right of the image uploaded, to actually see the icon, and delete the file. 

I've changed the color, to make it always visible when the file is uploaded. 

UPDATE : I've put the default color in RED (After Pierre's discussions)  - to be aligned between passenger / driver colors  + IconColor can be chosen :) 

[DAB-1169]: https://retool.atlassian.net/browse/DAB-1169?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Uploader/styles.js
+++ b/src/components/Uploader/styles.js
@@ -85,7 +85,9 @@ export const PreviewActions = styled.div`
   z-index: 15;
 `;
 
-export const TrashButton = styled(IconButton)`
+export const TrashButton = styled(IconButton).attrs({
+  icon: IconButton.ICONS.IconTrash,
+})`
   z-index: 20;
   margin: ${theme.space.m};
   align-self: end;
--- a/src/components/Uploader/styles.js
+++ b/src/components/Uploader/styles.js
@@ -2,6 +2,7 @@ import styled from 'styled-components/macro';
 import FormElementWrapper from '../FormComponents/FormElementWrapper';
 import Icon from '../Icon';
 import UiText from '../UiText';
+import IconButton from '../IconButton';
 
 import { theme } from '../../theme';
 import { STATES } from './constants';
@@ -79,12 +80,20 @@ export const PreviewActions = styled.div`
   bottom: -1px;
   border-radius: ${theme.borderRadius.s};
   background: ${theme.color.element.overlay};
-  padding: 0 ${theme.space.m};
-  text-align: right;
-  opacity: 0;
   pointer-events: none;
+  opacity: 0;
   z-index: 15;
-  transition: opacity 0.1s ease-out;
+`;
+
+export const TrashButton = styled(IconButton)`
+  z-index: 20;
+  margin: ${theme.space.m};
+  align-self: end;
+  padding: ${theme.space.s} !important;
+  background-color: #ffffff !important;
+  * {
+    color: ${({ iconColor }) => iconColor || theme.color_v3.feedback.error};
+  }
 `;
 
 export const Container = styled(FormElementWrapper)`
@@ -96,7 +105,6 @@ export const Container = styled(FormElementWrapper)`
   border-radius: ${theme.borderRadius.s};
   color: ${styles.textColor};
   background-color: ${theme.color.element.primary};
-  transition: all 0.15s ease-out;
   border: 1px dashed ${styles.borderColor};
 
   &:hover {
--- a/src/components/Uploader/index.js
+++ b/src/components/Uploader/index.js
@@ -2,7 +2,6 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 import Icon from '../Icon';
-import IconButton from '../IconButton';
 import Spinner from '../Spinner';
 import UploaderImageItem from '../UploaderImageItem';
 import UploaderItem from '../UploaderItem';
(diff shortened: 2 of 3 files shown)


src/components/Uploader/styles.js
```diff
@@ -81,10 +81,12 @@ export const PreviewActions = styled.div`
   background: ${theme.color.element.overlay};
   padding: 0 ${theme.space.m};
   text-align: right;
-  opacity: 0;
```
ganlhi, line 84: https://user-images.githubusercontent.com/512823/192714456-86dd046f-c48d-45af-b70a-715f559ed791.mov

That doesn't really work because now the whole thing is black when you don't have your mouse on it. I would make the background transparent there, and add a rule to make it black on hover. And maybe the `transition` rule here is not useful anymore.

ganlhi (changes requested):

src/components/Uploader/styles.js
agathepey: I'm stupid I was using a black screenshot to do my test, I haven't noticed 🤦‍♀️ changing now

ganlhi: That's an original reason to not see a problem 😄

keanan: The way this component was designed was to show the bin icon on hover, we should be able to make the button white on hover:
image

keanan: It's probably a lot nicer to change the design to always show the bin icon (not only on hover) what do you guys think? - to do that we would need to add a dark background to the button and make the icon white, so that it is visible on any images dark/light.

agathepey: I've asked Pierre, he told me to keep it blue  🤷‍♀️
What do you think about this version? The trash icon will always be visible

keanan: I think that is great but as mentioned before, if we always show the button we will need to add a background to the button as it will not display nicely on darker images.
image

keanan: Something like this, we can maybe check with Pierre what colors he prefers:
image

agathepey: On my opinion, it doesn't look great with a background 😝 I'm sending the different pictures to Pierre to let him choose the best option

agathepey: Pierre prefers with a white background, i'm changing it :)

keanan: But that will be bad UX - what if the image that is being uploaded is the same blue color? we will not see the icon. 
Maybe Pierre can come up with a nicer background color, maybe one with a bit of transparency.

agathepey: Yes true... Let's goo with a background

src/components/Uploader/styles.js
```diff
@@ -79,12 +80,18 @@ export const PreviewActions = styled.div`
...
-  text-align: right;
-  opacity: 0;
   pointer-events: none;
+  opacity: 0;
   z-index: 15;
-  transition: opacity 0.1s ease-out;
+`;
+
+export const TrashButton = styled(IconButton)`
+  color: ${theme.color.brand.secondary};
```
keanan, line 89: Please let's also allow for other colors to be chosen.

keanan:
Hey @agathepey please take a moment with Pierre to QA your PR in storybook - the hover colors and padding (multiple files) need some ✨ 

Also please see Guilhem's comment on the transition rule.

keanan:
Hey @agathepey what is happening with this PR? You will need to rebase as Guilhem has already merged another PR

keanan (approved):

agathepey:
Hey ! It's now ready to be merged, for the record, after discussions, Pierre chose to change the icon trash button to RED by default (to be aligned with passenger & driver designs)  - but the component accepts others colors, to be able for us to have it in blue easily :)

src/components/Uploader/styles.js
```diff
@@ -79,12 +80,20 @@ export const PreviewActions = styled.div`
...
+export const TrashButton = styled(IconButton)`
+  z-index: 20;
+  margin: ${theme.space.m};
+  align-self: end;
+  padding: ${theme.space.s} !important;
+  background-color: #ffffff !important;
+  * {
+    color: ${({ iconColor }) => iconColor || theme.color_v3.feedback.error};
+  }
 `;
```
ganlhi, line 97: 🌷 Just so you know, as this is a trash button, you could include the icon in it's definition so that the caller doesn't have to set it. 
```suggestion
export const TrashButton = styled(IconButton).attrs({ icon: IconButton.ICONS.IconTrash })`
  z-index: 20;
  margin: ${theme.space.m};
  align-self: end;
  padding: ${theme.space.s} !important;
  background-color: #ffffff !important;
  * {
    color: ${({ iconColor }) => iconColor || theme.color_v3.feedback.error};
  }
`;
```

ganlhi (approved):

src/components/Uploader/styles.js
agathepey: Not sure I understand your point here 🙈  Is it better practice to set up the icon in style rather than in the component itself in index.js ?

ganlhi: Not always but in this case, will you ever call `TrashButton` with another icon than the trashcan? Probably not, so it makes sense to make the icon baked in the component, as a default value, so that when you use `TrashButton` all you really need to provide is an `onClick` method.

agathepey: Ok thanks for the explanation, I change it with the icon directly setup in style :)

agathepey: Ok for u @ganlhi ? :)

ganlhi: Yes, looks good 👍🏻

PR #247, opened 2022-10-04
Improve styling

ganlhi:
# `Toggle` component

Aligned text style of the label with other fields. 

Before:
image

After:
image

# `Select` component

Improved styling and behaviour of the blank option/placeholder.

Before:
image
image

After:
image
image

There is always an empty option, and it is selected if no `defaultValue` is provided. But if no placeholder text is provided for this empty option, the option itself doesn't appear in the list. 
If a placeholder is provided, it is properly styled, and the option in the list is not selectable once another value has been selected.

# Misc.

Set the default placeholder text color to an actual color of the palette instead of the browser's default.

### DIFF
--- a/src/components/Select/index.js
+++ b/src/components/Select/index.js
@@ -1,29 +1,49 @@
-import React from 'react';
+import React, { useCallback, useState } from 'react';
 import PropTypes from 'prop-types';
-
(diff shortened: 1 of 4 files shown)


keanan (approved):
✨

PR #258, opened 2023-03-16
[DAB-1528] Small fixes on components

agathepey:
- [X] **Card** without Text element add font-family default + useless style
- [X] **Item** - valueIcon to accept null for no Icon + accept now borderColor
- [x] **Input** -  Delete !important style + allow borderColor 
- [X] **InputFiled** - increase label bottom margin 
- [X] **Tabs** - accept fontSize
- [X] **Button** - Delete !important style + `disableAutoMarginLeft`
- [X] **Checkbox** - Delete !important style
- [X] **FormElement** - Delete !important style
- [X] **IconButton** -  swap {...props} under style + Delete !important style + delete padding
- [X] **Radio** - Delete !important style
- [X] **Selector** - Delete !important style
- [X] **Uploader** - Delete !important style
- [X] **TextArea** - Delete !important style

**TablePagination**:
- [x]  Change `disableColSort` to `disableSortBy`
- [x]  Fix `TypeError: Cannot read properties of undefined (reading 'length')` on playground

**Table changes**:
- [x]  Add colWidth - (same as Table Pagination) 
- [x]  Add disableSortBy - (same as Table Pagination) && remove chevron if disabled
- [x]  Table on row hover #primaryInactive

**DatepickerDay** 
- [X]  react-datepicker version updated to `4.10`
- [X]  Accept now color flexible with `datepickerColor`
- [X] New dropdown style


src/components/Button/styles.js
ganlhi, line 1: Could we take advantage of this PR to add something (maybe a prop) to easily disable [this rule](https://github.com/heetch/flamingo/blob/66858ea9314effcb742b207b36d474712f66646e/src/components/Button/styles.js#L156)? When we stack buttons vertically, this is very annoying!

src/components/DatepickerDay/index.js
```diff
@@ -21,6 +21,7 @@ const DatepickerDay = ({
   dateFormat,
   locale,
   withPortal,
+  isPink,
```
ganlhi, line 24: Can't we take advantage of the `iconColor` I added before to implement this in an even more flexible way?

agathepey: Yes you r right, will be better to have something more flexible 
About that, Keanan wanted to discuss this afternoon during our frontend meeting about adding a rule to change color easily for all the flamingo components (because it's always annoying when we want a component with primary or secondary styles), as it's related I leave it as it's for now :)

src/components/DatepickerRange/index.stories.js
```diff
@@ -0,0 +1,101 @@
...
+    <DatepickerRange
+      id='default'
+      value={new Date('2020-08-04')}
+      isClearable
+      startPlaceholder='Start date range'
+      endPlaceholder='End date range'
+      isPink
+    />
+    <Code>{`
+    <DatepickerDay id='default' value={new Date('2020-08-04')} isClearable startPlaceholder='Start date range' endPlaceholder='End date range' isBlue/>
```
ganlhi, line 44:
```suggestion
    <DatepickerDay id='default' value={new Date('2020-08-04')} isClearable startPlaceholder='Start date range' endPlaceholder='End date range' isPink/>
```

keanan: Would it not be better to use brand= 'primary` or `secondary` so that in the future if the colors change we can keep the same naming?

ganlhi: As I said on Friday, I would even go for something wider. Let's pretend it's in TypeScript already, we could have a prop named `accentColor`:
```typescript
{
  accentColor: 
    | { brand: 'primary' | 'secondary' | 'tertiary' }
    | { color: string }
}
```
For the branded variant, we could even go further and allow most of the variants we have [here](https://flamingo.pages.dev/?path=/story/design-system-colors--interface-colors) and [here](https://flamingo.pages.dev/?path=/story/design-system-colors--brand-colors). For instance things like `passenger-primary`, `driver-primary`, `passenger-secondary`, `driver-secondary`, or `brand-red-3`.

ganlhi: But this is maybe a little ambitious for now and we could wait for a bigger refactor to implement something like that 😅

agathepey: Yes I guess it will be best to have a separate PR for that

src/components/Button/styles.js
agathepey: I've deleted the line, but please be aware, that this can change actual design we have for button with a margin

agathepey: Hey @ganlhi :) 
So you would like to have a prop, something like `disableAutoMarginLeft` to be able to not have a marginLeft for the button if needed ? Just make sure I understood what you want, thank you !

ganlhi: Yes, maybe something like that. Ideally it would be the reverse, but that would break existing code...

agathepey: Let me know what you think, I choose a solution without having to change the existing code to be easier/faster to implement this new version :)

src/components/Button/styles.js
```diff
@@ -78,8 +78,20 @@ const styles = {
...
   },
+  marginLeft({ margin, disableAutoMarginLeft }) {
+    if (disableAutoMarginLeft) {
+      return 0;
+    }
+    if (!margin) {
+      return theme.space.m;
+    }
+    return undefined;
+  },
```
ganlhi, line 89: Looks good for the disabled version. But I see you set a default `space.m` value if no margin is specified, but you never use this `margin` param if set. Why?

agathepey: `margin` props is used on line 145. I kept the logic as it was, to only reset the margin-left for the second button that needs margin-left, let me know if it's unclear :)

src/components/DatepickerDay/index.js
```diff
@@ -44,6 +45,7 @@ const DatepickerDay = ({
         invalid={invalid}
         valid={valid}
         autoFocus={stateHasFocus}
+        borderColor={datepickerColor && datepickerColor}
```
keanan, line 48: Is this prop being used? If so please can you add a default color (same as you did with RangePIcker)

src/components/DatepickerRange/index.js
```diff
@@ -0,0 +1,178 @@
...
+      <StyledDatepickerInput
+        {...dateInputProps}
+        ref={ref}
+        placeholder={placeholder}
+        icon={Icon.ICONS.IconCalendar}
+        iconLeft={iconLeft}
+        invalid={invalid}
+        valid={valid}
+        borderColor={datepickerColor && theme.color.brand.primary}
+        {...inputProps}
```
keanan, line 49: If there is only 1 input being used twice (with a single `inputProps`) how can the user define id's (or other unique props) for each input?

keanan (changes requested):
Thanks for doing this for us. I have made some small comments and suggestions. 

The main concern is for the DatePickerRange as we are only sending 1 set of input props but using 2 inputs (start & end) - have you maybe tried to test this on a form where you would need to register the id's and get the values for start/end date?

ganlhi (approved):

src/components/DatepickerRange/index.js
agathepey: Do you think we will have a situation where we want different props for start and end date picker ? Like a different icon or bordercolor ? (expect Id, which needs to be different)
 Do you want me to have all the props for end and start like the startPlaceholder, so we will have `startIconLeft` `endIconLeft` and so on ?

keanan: How will the user get `start_date` and `end_date` back from the Flamingo component? Have you tested this with a form to see you are getting the values back?

ganlhi: Agathe, maybe the icons, in case we want a different icon for each. But the colors (of the icon, the border...) I don't think so. I don't see why in a design we would want different colors of fields within the same form.

keanan (approved):

